Issue #56, opened 2026-04-27
Extend fairness visualization to support multiple sensitive features


Khanz9664:
# [ENHANCEMENT] Support multiple sensitive features in fairness visualization

### 📌 Objective
Extend fairness visualization to support **multiple sensitive features** instead of handling only one feature at a time.

---

### Problem
Currently, fairness plots effectively use only the first sensitive feature when multiple are provided.

This leads to:
- incomplete visualization coverage
- hidden subgroup disparities
- inconsistent behavior for multi-feature inputs

---

### Goal
Enable the system to:
- generate plots for **each sensitive feature**
- clearly label outputs per feature
- optionally return structured outputs (e.g., `{feature_name: figure}`)

---

### Relevant Files
- `trustlens/visualization/fairness.py`
- `trustlens/visualization/__init__.py`

---

### Implementation Notes
- Maintain backward compatibility
- Avoid silent dropping of features
- Ensure plots are clearly labeled
- Integrate cleanly with `_plot_bias()`

---

### Acceptance Criteria
- Multiple sensitive features are supported
- No feature is silently ignored
- Output is clear and structured
- Existing API behavior remains unchanged

---

### 🤝 Want to work on this?
Comment below and feel free to pick it up!

cc: @komoike-oss28-ui

komoike-oss28-ui:
Hi @Khanz9664! It's been a while — I previously contributed to this project and am happy to be back.
I'd like to take on this issue. Since I worked on fairness.py recently, I have a good understanding of the existing structure and can implement multi-feature support while keeping backward compatibility intact.
Could you assign this to me? I'll get started right away.

Khanz9664:
Welcome back @komoike-oss28-ui — great to have you again! 

Thanks for picking this up, this is a really valuable improvement to the fairness module.

Feel free to proceed 👍

A couple of guidelines to keep in mind:
- Maintain backward compatibility with the current API
- Avoid silently dropping any sensitive features
- Prefer a clear and structured output (e.g., per-feature plots or grouped outputs)
- Keep integration clean with the existing plot_bias() flow

No need to over-engineer the first version — we can iterate on the API if needed.

Looking forward to your PR!

PR #76, opened 2026-05-06
Feat: add multi_feature parameter to TrustReport.plot_bias() (#74)

### LINKED ISSUE #74: [FEATURE]: Add multi_feature parameter to TrustReport.plot_bias() for per-feature visualization output
### Problem Statement

`TrustReport.plot_bias()` currently does not expose multi-feature outputs at the API level. When multiple sensitive features are provided to `analyze()`, users have no way to retrieve per-feature visualizations through `plot_bias()` itself — they must drop down to the lower-level `*_multi` wrappers added in #56 directly.

After the pipeline-level integration in #70, the underlying `_plot_bias()` already handles multiple features cleanly. The remaining gap is exposing this capability through the public `plot_bias()` API in a way that is opt-in and does not change existing behavior.

### Proposed Solution

Add an opt-in `multi_feature: bool = False` parameter to `TrustReport.plot_bias()`. `mode` and `multi_feature` are treated as independent dimensions, and the return shape is structured (no flattening):

| `mode` | `multi_feature` | Return type |
|---|---|---|
| single (`"subgroup"`, `"equalized_odds"`, `"gap"`, `"summary"`) | `False` (default) | `Figure` (current behavior) |
| single | `True` | `dict[str, Figure]` keyed by feature name |
| `"all"` | `False` (default) | `dict[str, Figure]` keyed by mode (current behavior) |
| `"all"` | `True` | `dict[str, dict[str, Figure]]` keyed by mode, then by feature name |

**Backward compatibility:** strictly preserved. Default behavior (`multi_feature=False`) is unchanged; all existing call sites continue to work without modification.

**Implementation approach:**
- Reuse `_plot_bias()` output rather than reimplementing plotting logic
- `plot_bias()` acts as a thin transformation layer over that output
- Visualization-only — no file saving
- Deterministic feature ordering (matches existing pipeline ordering)

**Example usage:**

​```python
**Default — unchanged**
fig = report.plot_bias()

**Single mode, multi-feature**
figs = report.plot_bias(mode="equalized_odds", multi_feature=True)
{"gender": Figure, "age": Figure}

**All modes, multi-feature**
nested = report.plot_bias(mode="all", multi_feature=True)
{"subgroup": {"gender": Figure, "age": Figure},
"equalized_odds": {"gender": Figure, "age": Figure},
"gap": {"gender": Figure, "age": Figure}}
​```

**Tasks:**
- [ ] Add `multi_feature: bool = False` parameter to `TrustReport.plot_bias()`
- [ ] Implement the transformation layer over `_plot_bias()` output
- [ ] Add unit tests covering all four cells of the return-shape table
- [ ] Update docstring with the new parameter and return shapes
- [ ] Update CHANGELOG

### Alternatives Considered

- **Combined subplot layout** (e.g., putting all features into a single figure with subplots): Considered as a UX improvement but rejected for this issue because it conflates two concerns. The structured `dict[str, Figure]` output gives users full flexibility to compose their own layouts. Combined subplot layouts can be addressed in a separate, more design-focused follow-up.
- **Flattened keys** (e.g., `{"subgroup__gender": Figure, ...}`): Rejected because it loses the `mode` vs `feature` semantic separation that mirrors the rest of the API.
- **A separate method** like `plot_bias_multi()`: Rejected to avoid API duplication and keep `plot_bias()` as the single entry point for bias visualization.

### Additional Context

Follow-up to #56 (multi-feature wrappers in `fairness.py`) and #70 (pipeline-level `plot_module` / `_plot_bias` integration).

Design was discussed and aligned with @Khanz9664 in the #56 PR thread before this issue was opened.

komoike-oss28-ui:
## Summary

Adds an opt-in `multi_feature: bool = False` parameter to `TrustReport.plot_bias()` so callers can retrieve per-feature visualizations through the public API instead of dropping down to the lower-level `plot_*_multi` helpers added in #56.

`mode` and `multi_feature` are independent dimensions; the return shape is fully determined by their combination, with no flattening and no `None` values in multi-feature outputs.

| `mode` | `multi_feature` | Return type |
| --- | --- | --- |
| single (`summary`, `subgroup`, `equalized_odds`, `gap`) | `False` (default) | `Figure` (current behavior, unchanged) |
| `summary` | `True` | `Figure` (passthrough — same as `False`) |
| `subgroup` / `equalized_odds` / `gap` | `True` | `dict[str, Figure]` keyed by feature |
| `all` | `False` (default) | `dict[str, Figure \| None]` (current behavior, unchanged) |
| `all` | `True` | `dict[str, dict[str, Figure]]` keyed by mode then feature |

## Related Issue

Closes #74.

Design was locked down with @Khanz9664 in the issue thread before implementation.

## Type of Change

Please check the type of change:
- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [x] 🚀 New feature (non-breaking change which adds functionality)
- [ ] 📝 Documentation update (changes to docs or examples)
- [ ] 🧹 Maintenance (refactoring, code cleanup, dependencies)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Changes Made

- [x] Added `multi_feature: bool = False` parameter to `TrustReport.plot_bias()` in `trustlens/report.py`. Implemented as a thin transformation layer over the existing `plot_subgroup_performance_multi`, `plot_equalized_odds_multi`, and `plot_fairness_gap_multi` helpers — no plotting logic is reimplemented.
- [x] `mode="all", multi_feature=True` always returns the three keys `subgroup → equalized_odds → gap` in that order. Components with no underlying data map to `{}` (empty dict), never `None`.
- [x] `summary + multi_feature=True` falls through to the existing single-figure path (no per-feature expansion for summary, as agreed).
- [x] Feature ordering uses `sorted(feature_names)` via a `_sorted_feature_dict()` helper, so output is deterministic regardless of insertion order.
- [x] `'gap'` keeps the same priority rule as the single-feature path (`subgroup_performance` first, falling back to `equalized_odds`).
- [x] When `multi_feature=True`, `save_path` is ignored with a `logger.warning`. Per-feature file saving is intentionally not exposed here; callers who need on-disk output can use the `plot_*_multi` helpers directly.
- [x] Updated docstring with the new return-shape table and parameter description.
- [x] Added `tests/test_plot_bias_multi_feature.py` (18 cases across 6 test classes).
- [x] Updated `CHANGELOG.md` under `[Unreleased] / Added`.

Default behavior (`multi_feature=False`) is strictly preserved — the `mode="all", multi_feature=False` branch is byte-for-byte unchanged.

## Testing

```
make test       → 237 passed (219 pre-existing + 18 new)
make lint       → All checks passed
pre-commit run  → all hooks passed (ruff, ruff-format, mypy)
```

`tests/test_plot_bias_multi_feature.py` covers:

- `TestReturnShapeMatrix` — all four `(mode, multi_feature)` cells of the return-shape table.
- `TestSummaryMultiFeaturePassthrough` — `summary + multi_feature=True` returns the same type as `multi_feature=False`.
- `TestPartialDataHandling` — only `subgroup_performance`, only `equalized_odds`, no `bias` module at all, `'gap'` fallback when `subgroup_performance` is absent, and an explicit assertion that no value at any level is `None`.
- `TestDeterministicOrdering` — mode key order, sorted feature order (using a fixture that inserts `gender` before `age` to make the sort observable), and stability across repeated calls.
- `TestInvalidModeWithMultiFeature` — `mode="bogus"` raises `ValueError` regardless of `multi_feature`.
- `TestBackwardCompatibility` — default call and explicit `multi_feature=False` produce identical types.

- [x] All unit tests pass locally.
- [x] Added new tests for the changes made.

## Pre-submit Checklist

Before submitting, please ensure you have:
- [x] Run `pre-commit run --all-files` and all hooks passed.
- [x] Verified that all tests pass (`make test`).
- [x] Updated the `CHANGELOG.md` if applicable.
- [ ] Verified that documentation has been updated for new features.

> Note on the docs checkbox: the user-facing docstring on `plot_bias()` is updated with the full `(mode, multi_feature)` return-shape table and parameter description. I did not touch `docs/metrics/bias.md` or `README.md` in this PR — happy to add a short usage example to either if you'd like it as part of this change.

## Notes for Reviewers

A few choices worth flagging for review:

1. **`save_path` behavior under `multi_feature=True`** — currently ignored with a `logger.warning`. The reasoning is that `plot_*_multi(save_dir=...)` already provides per-feature file saving with its own naming convention (`subgroup_performance_<feature>.png`, etc.), and overloading `plot_bias(save_path=...)` to multi would either need a new naming scheme or risk collisions. Happy to switch to raising `ValueError` instead if you'd prefer a stricter contract.

2. **Asymmetry between `multi_feature=False` and `multi_feature=True` for `mode="all"`** — the existing `multi_feature=False` path can return `None` values per the original contract; the new `multi_feature=True` path uses `{}` instead, per the spec we agreed on. The asymmetry is intentional (preserving backward compatibility), but I wanted to flag it explicitly so it's not surprising during review.

3. **`'gap'` priority rule** — mirrors the existing single-feature path (`subgroup_performance` first, then `equalized_odds`). If you'd rather have multi-feature `'gap'` always use `equalized_odds` when both are present (since `plot_fairness_gap()` is built on equalized-odds-shaped data), I'm happy to switch. I went with consistency between the two paths as the safer default.

### Out of scope (potential follow-ups)

- Combined subplot layout (per the alternatives discussion in the issue).
- Multi-figure support inside `plot_module()` itself — `_plot_bias()` currently returns a single figure and the dispatcher's contract is one figure per module. Lifting that would touch `plot_module()`'s shape and is better handled separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `_BIAS_PLOT_TYPES` — internal registry for deterministic plot-type dispatch ordering in `_plot_bias()`.
 - `tests/conftest.py` — centralized Agg backend configuration for the test suite.
 - `tests/test_plot_module_multi_feature.py` — 23 integration tests covering nested figure outputs, filename sanitization, orchestrated saving, and edge cases.
+- `TrustReport.plot_bias()` now accepts an opt-in `multi_feature: bool = False` parameter for per-feature visualization output. With `multi_feature=True`, single modes (`"subgroup"`, `"equalized_odds"`, `"gap"`) return `dict[str, Figure]` keyed by feature name, and `mode="all"` returns a nested `dict[str, dict[str, Figure]]` keyed by mode then feature. The structure is fixed by the `(mode, multi_feature)` combination, missing components are represented by empty dicts (never `None`), and feature ordering is deterministic (`sorted(feature_names)`). Default behavior (`multi_feature=False`) is unchanged. `tests/test_plot_bias_multi_feature.py` adds 18 tests covering the four return-shape cells, partial-data handling, deterministic ordering, and invalid-mode interaction (closes #74). Thanks @komoike-oss28-ui
 
 ### Improved
 - Final Trust Score logic now includes a base score, penalty breakdown, and decisive deployment verdicts.
--- a/tests/test_plot_bias_multi_feature.py
+++ b/tests/test_plot_bias_multi_feature.py
@@ -1,0 +1,296 @@
+"""
+tests/test_plot_bias_multi_feature.py.
+======================================
+Tests for the ``multi_feature`` parameter of ``TrustReport.plot_bias()``.
+
+These tests cover the four cells of the ``(mode, multi_feature)`` return-shape
+table, partial-data handling, deterministic ordering, and the interaction
+between ``multi_feature=True`` and an invalid ``mode``.
+"""
+
+from unittest.mock import MagicMock
+
+import matplotlib
+
+matplotlib.use("Agg")
+import matplotlib.pyplot as plt
+import numpy as np
+import pytest
+
+from trustlens.report import TrustReport
+
+# ---------------------------------------------------------------------------
+# Fixtures
+# ---------------------------------------------------------------------------
+
+
+def _make_report(bias_data: dict) -> TrustReport:
+    """Build a minimal TrustReport whose bias module holds ``bias_data``."""
+    model = MagicMock()
+    X = np.zeros((10, 2))
+    y_true = np.array([0, 1] * 5)
+    y_pred = np.array([0, 1] * 5)
+    y_prob = np.random.rand(10, 2)
+    return TrustReport({"bias": bias_data}, model, X, y_true, y_pred, y_prob)
+
+
+@pytest.fixture
+def multi_feature_bias_data() -> dict:
+    """Bias data with two sensitive features (gender, age) on both modes."""
+    return {
+        "subgroup_performance": {
+            "gender": {
+                "male": {"accuracy": 0.82, "f1": 0.80},
+                "female": {"accuracy": 0.70, "f1": 0.68},
+                "__summary__": {"performance_gap": 0.12},
+            },
+            "age": {
+                "young": {"accuracy": 0.79, "f1": 0.77},
+                "old": {"accuracy": 0.65, "f1": 0.63},
+                "__summary__": {"performance_gap": 0.14},
+            },
+        },
+        "equalized_odds": {
+            "gender": {
+                "male": {"tpr": 0.80, "fpr": 0.10},
+                "female": {"tpr": 0.65, "fpr": 0.20},
+                "__summary__": {
+                    "tpr_gap": 0.15,
+                    "fpr_gap": 0.10,
+                    "tpr_violation": "moderate",
+                    "fpr_violation": "acceptable",
+                },
+            },
+            "age": {
+                "young": {"tpr": 0.78, "fpr": 0.12},
+                "old": {"tpr": 0.60, "fpr": 0.25},
+                "__summary__": {
+                    "tpr_gap": 0.18,
+                    "fpr_gap": 0.13,
+                    "tpr_violation": "moderate",
+                    "fpr_violation": "moderate",
+                },
+            },
+        },
+    }
+
+
+@pytest.fixture
+def multi_feature_report(multi_feature_bias_data) -> TrustReport:
+    return _make_report(multi_feature_bias_data)
+
+
+# ---------------------------------------------------------------------------
+# 1. Return-shape matrix: four (mode, multi_feature) cells
+# ---------------------------------------------------------------------------
+
+
+class TestReturnShapeMatrix:
+    """Cover the four cells of the (mode, multi_feature) return-shape table."""
+
+    def test_single_mode_multi_feature_false_returns_figure(self, multi_feature_report):
+        """Single mode + multi_feature=False -> Figure (current behavior)."""
+        for mode in ("subgroup", "equalized_odds", "gap"):
+            fig = multi_feature_report.plot_bias(mode=mode, multi_feature=False, show=False)
+            assert isinstance(fig, plt.Figure), f"mode={mode} did not return a Figure"
+
+    def test_single_mode_multi_feature_true_returns_dict_of_figures(self, multi_feature_report):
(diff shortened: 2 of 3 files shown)


Khanz9664 (changes requested):
Excellent work on this implementation @komoike-oss28-ui — I’ve completed a full local review and verification of the PR, and everything looks solid 👍

Key things verified during review:

* strict adherence to the design/spec discussed in #74
* deterministic ordering guarantees implemented correctly
* clean structured return shapes for all `(mode, multi_feature)` combinations
* proper handling of partial/missing data using `{}` instead of `None`
* full backward compatibility preserved for existing users
* no duplicated plotting logic — implementation cleanly reuses the existing `*_multi` wrappers
* comprehensive edge-case test coverage (all 237 tests passing locally)

The updated docstring + return-shape matrix are also very well done and significantly improve the developer experience around the API.

I only have one very small cleanup request before merge:

* please remove the local:

  ```python
  matplotlib.use("Agg")
  ```

  from `tests/test_plot_bias_multi_feature.py`

We already centralized backend configuration in `tests/conftest.py` during the visualization backend cleanup, so this is now redundant.

Other than that, this is merge-ready from my side.

komoike-oss28-ui:
Thanks for the thorough review @Khanz9664 — really appreciate the detailed verification 🙏

Done. Removed the local `matplotlib.use("Agg")` from `tests/test_plot_bias_multi_feature.py` in [`d1e550c`](commit hash will auto-link after push). Confirmed:

- `tests/conftest.py` already sets the backend globally
- All 18 new tests still pass with the local call removed
- `make test` → 237 passed
- `pre-commit run` → all hooks passed

The import block now matches the cleaner style used in `test_visualization_api.py` (just `import matplotlib.pyplot as plt`). Ready for merge whenever you are 👍

Khanz9664 (approved):
Perfect — thanks for the quick follow-up @komoike-oss28-ui 👍

The backend setup is now fully aligned with the centralized `tests/conftest.py` approach, and the cleanup keeps the test suite architecture much more consistent going forward.

I re-checked the updated changes and everything looks good:

* deterministic return contracts preserved
* backward compatibility intact
* multi-feature API behavior stable
* tests and hooks all passing cleanly

This is ready to merge.

komoike-oss28-ui:
Thanks @Khanz9664 for the careful review and guidance throughout — appreciated the design discussion in #74 as well, it made the implementation much cleaner. Happy to contribute again 👍

Khanz9664:
Really appreciate the contribution @komoike-oss28-ui — this was a great addition to the project 🙌
Feel free to pick up another issue if something looks interesting.
_Looking forward to more contributions from you._

Issue #57, opened 2026-04-27
Create a centralized styling system for visualization's.


Khanz9664:
# [ENHANCEMENT] Centralize visualization styling (palette, themes, defaults)

### Objective
Create a centralized styling system for all visualization modules.

---

### Problem
Styling (palette, themes, grid settings) is currently defined inside individual plotting functions.

This leads to:
- duplicated code
- inconsistent visuals
- difficulty maintaining a unified look

---

### Goal
Introduce a shared styling configuration that:
- defines global color palettes
- applies consistent themes (matplotlib/seaborn)
- standardizes plot appearance

---

### Suggested Location
`trustlens/visualization/style.py`

---

### Implementation Notes
- Move styling logic into reusable functions
- Refactor existing plots to use centralized styling
- Keep flexibility for future customization

---

### Acceptance Criteria
- No hardcoded styles in plotting functions
- All plots share a consistent look
- Styling can be updated from one place
- No breaking changes

---

### 🤝 Want to work on this?
Comment below and feel free to pick it up!

cc: @komoike-oss28-ui

komoike-oss28-ui:
@Khanz9664, it’s been a while, but would it be okay if I took on this project?🙏

Since it’s been a while since we last spoke, I’d like to go over the design before I start coding. I have a few questions:

1. **Scope of modules**: Should the centralization cover all of `visualization/*` (calibration, failure, bias, fairness, representation), or do you want to start with a subset as a proof of concept? I'd lean toward covering all of them in one PR for consistency, but happy to split if you prefer smaller reviewable chunks.

2. **API shape for `style.py`**: A few options I see —
   - (a) A `set_style()` / `apply_style()` function called once at plot entry (simplest, mirrors `sns.set_theme()` patterns)
   - (b) A constants module exposing `PALETTE`, `GRID_KWARGS`, `FIG_DEFAULTS` dicts that each `_plot_*` function unpacks
   - (c) A context manager (`with trustlens_style(): ...`) so plots don't leak rcParams globally
   
   I'd default to (b) + a small helper in (a) for the rcParams side, since it stays explicit and avoids global state surprises. Any preference?

3. **Themes**: Do you want a single default theme for now, or should the design support multiple named themes (e.g. `"default"`, `"dark"`, `"colorblind"`) from day one? Either is fine — just want to know how extensible to make the entry point.

[...]

PR #102, opened 2026-05-19
feat(visualization): centralize plotting style — draft for review (#57)

komoike-oss28-ui:
## Summary

First migration step toward centralizing visualization styling per #57. Introduces an internal `style.py` module and refactors `calibration_plots.py` as a representative test case. Opening as a draft so the abstraction can be sanity-checked before the remaining modules (`bias_plots`, `failure_plots`, `fairness`, `representation_plots`, `summary_plot`) are migrated on this same branch.

**Key change**: `matplotlib.rcParams` are now scoped via an `apply_style()` context manager instead of mutated globally — addresses the notebook/pipeline concern raised on the issue.

## Related Issue

Relates to #57

(This PR migrates `calibration_plots.py` only; remaining visualization modules will follow as additional commits on the same branch. The issue will be closed once all modules are migrated.)

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [x] 🚀 New feature (non-breaking change which adds functionality)
- [ ] 📝 Documentation update (changes to docs or examples)
- [x] 🧹 Maintenance (refactoring, code cleanup, dependencies)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Changes Made

- [x] Added internal `trustlens/visualization/style.py` — single source of truth for `BRAND_COLORS`, `PALETTE`, `SEMANTIC_COLORS` (`severity` / `verdict` / `grade` / `direction` / `neutral`), `TYPOGRAPHY`, `GRID`, `FIG_DEFAULTS`, `SPACING`, plus a frozen `Theme` dataclass
- [x] Added `apply_style()` context manager that scopes `matplotlib.rcParams` mutations to a `with` block, restoring previous state on exit
- [x] Added `styled_figure()` and `get_categorical_colors()` helpers
- [x] Refactored `calibration_plots.py` to source colors, grid alpha, and savefig DPI from the centralized theme (visual output unchanged)
- [x] Added `tests/test_visualization_style.py` (34 tests) covering color contracts, `Theme` behavior, rcParams restoration (normal exit / exception / nested), and helpers
- [x] Updated `CHANGELOG.md` under `[Unreleased] / Added`
- [ ] Still to do on this branch: migrate `bias_plots`, `failure_plots`, `fairness`, `representation_plots`, `summary_plot`

## Testing

​```
$ pytest tests/
============================= 297 passed in 10.08s =============================

$ pre-commit run --all-files
... (all hooks passed: trailing-whitespace, eof-fixer, ruff, ruff-format, mypy)
​```

**Visual parity**: same hex values, figsize, alpha, and linewidth as before. The only behavioral change is that `matplotlib.rcParams` are no longer permanently mutated after a plot call.

- [x] All unit tests pass locally.
- [x] Added new tests for the changes made.

## Pre-submit Checklist

- [x] Run `pre-commit run --all-files` and all hooks passed.
- [x] Verified that all tests pass (`make test`).
- [x] Updated the `CHANGELOG.md` if applicable.
- [x] Verified that documentation has been updated for new features. *(N/A — module is internal.)*

## Notes for Reviewers

A few design judgments worth flagging:

1. **`apply_style()` is a `@contextmanager` function, not a class** — `try` / `finally` makes the rcParams restore explicit; no re-entry use case here.
2. **`Theme` is `frozen=True` with `default_factory` for mutable fields** — each instance gets its own palette/brand/semantic copies, so mutating one doesn't poison `DEFAULT_THEME` (tested).
3. **Did not migrate magic numbers (`lw=1.5`, `alpha=0.15`, `figsize=(7, 8)`)** — centralizing color and rcParams is high-leverage; lifting every plot-local constant would balloon `style.py` with single-use values. Happy to promote them later when a second module needs the same value.
4. **`SEMANTIC_COLORS["severity"]` keys**: `acceptable` / `moderate` / `severe` / `unknown` — matches `fairness.py`'s vocabulary plus an explicit fallback. Distinct from `verdict` (`deploy` / `caution` / `do_not_deploy`) and `grade` (`A`–`D`) per our discussion.
5. **Internal module — no `__all__`** per your call on #57. Promote to public once the shape stabilizes.
6. **Switched to `Relates to #57`** per your feedback — the issue stays open until all visualization modules are migrated on this branch. Commit messages still reference "Closes" in commit 3 (`b2a57f8`), but GitHub uses the PR-level link for auto-close, so the issue won't close prematurely on merge.
Most likely to want adjustment before the remaining migration: the `apply_style()` API shape and the `SEMANTIC_COLORS` key naming. Let me know what you'd tweak 🚀

## Summary by CodeRabbit

* **New Features**
  * Centralized, theme-driven visualization styling with a frozen default theme.
  * New helpers: a scoped style context manager, a themed figure helper, and a categorical-color utility.

* **Bug Fixes**
  * Plots now consistently use theme colors, grid, and DPI; show/save behavior made predictable.

* **Tests**
  * Added tests validating style constants, theme behavior/helpers, and plotting input validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Khanz9664/TrustLens/pull/102?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


komoike-oss28-ui:
Hi @Khanz9664, draft PR is up: #102 🙏

Quick summary of what's in it for sanity-check:
- New internal `trustlens/visualization/style.py` with `PALETTE`, `BRAND_COLORS`, `SEMANTIC_COLORS` (severity / verdict / grade / direction / neutral), typography / grid / figure defaults, and a frozen `Theme` dataclass
- `apply_style()` as a context manager — `rcParams` mutations are scoped to a `with` block, addressing the notebook/pipeline concern you raised
- `calibration_plots.py` migrated as the first proof-of-concept; visual output preserved (same hex / figsize / alpha)
- 34 tests for the new module, all 298 tests passing locally
- Remaining modules (`bias_plots`, `failure_plots`, `fairness`, `representation_plots`, `summary_plot`) intentionally left for follow-up commits on the same branch, pending your feedback on the abstraction shape

The PR description has the full design rationale and a couple of judgment calls flagged in **Notes for Reviewers** that are most likely to want adjustment (especially the `apply_style()` API shape and the `SEMANTIC_COLORS` key naming).

Also folded in three bot-flagged fixes (one of which surfaced a pre-existing `plt.show()` backend-check bug) into a follow-up commit on the same branch.

Whenever you've got a moment 🚀

Khanz9664 (commented):
@komoike-oss28-ui Thanks for putting this together — I spent some time going through the implementation, the tests, and the rationale in the PR description, and overall I think the direction is very solid. This feels like a thoughtful refactor rather than a “move constants around” PR, which is exactly what I was hoping for from this issue.

The biggest thing I wanted to sanity-check was the abstraction shape, and I think you landed in a good place. Keeping `style.py` internal for now, separating `BRAND_COLORS` from `SEMANTIC_COLORS`, using a frozen `Theme` dataclass, and especially scoping `rcParams` through a context manager instead of mutating global matplotlib state all make sense for where TrustLens is today. The notebook/pipeline concern was one of my biggest worries here, so I’m happy to see that handled cleanly.

I also appreciate the restraint in only migrating `calibration_plots.py` first rather than trying to refactor the entire visualization layer in one shot. That makes the architectural review much easier before the more opinionated modules (`summary_plot.py`, fairness, representation, etc.) get touched.

The test coverage is also excellent here. The rcParams restoration tests (especially nested/exception paths) are exactly the kind of thing that tends to get missed in visualization refactors, so thank you for being thorough there.

A couple of small things I’d still like addressed before we continue migrating the remaining modules:

**1. Style fallback warning**

I agree with Llama’s feedback here. I’m okay with keeping the fallback behavior (we definitely don’t want plotting to fail because a matplotlib install is missing a style sheet), but silently swallowing it feels a little too quiet for debugging. A lightweight `warnings.warn(...)` would probably be the right middle ground so custom theme failures are visible without becoming hard failures.

**2. `n_bins` validation**

CodeRabbit also raised a fair point here. Since width is derived from `n_bins`, I’d prefer we validate `n_bins > 0` explicitly and fail with a clear error message instead of letting a divide-by-zero or weird negative width happen downstream. Small change, but good defensive hardening.

**3. Migration strategy**

I’m good with the overall abstraction now, so once the above is addressed I think you can continue migrating the remaining visualization modules on this same branch. I’d especially keep an eye on `summary_plot.py` and `fairness.py` since they currently encode the most semantic color logic and are probably the easiest places for drift to creep in.

One thing I specifically *don’t* want us to lose during migration is visual parity. Keeping outputs effectively pixel-equivalent for this PR (same sizing, spacing, severity semantics, etc.) is still the right call — we can iterate on aesthetics separately once centralization lands.

Also, don’t close #57 yet—this PR is only the first migration step. Change Closes #57 to Partially addresses #57 or Relates to #57 until all visualization modules are migrated.

Overall though: really nice work so far. This is moving in the exact direction I had in mind for #57 🚀

komoike-oss28-ui:
Thanks for the review @Khanz9664 🙏 All three points addressed in `7100d23`:

1. **Style fallback warning** — `apply_style()` now emits a `UserWarning` when the base style sheet can't be loaded, instead of silently falling through. Custom theme typos will be visible in debugging without breaking plots.
2. **`n_bins` validation** — `plot_reliability_diagram()` now validates `n_bins >= 1` upfront with a clear `ValueError`, before any width calculation. Added two regression tests in a new `tests/test_calibration_visualization.py` (following the repo's per-module visualization test convention).
3. **`Closes #57` → `Relates to #57`** — updated the PR description; commit message in `b2a57f8` still says "Closes" but GitHub uses the PR-level link, so the issue won't auto-close on merge.

All 301 tests still passing locally, `pre-commit run --all-files` clean.

Will start on the remaining migrations next — beginning with `failure_plots.py` and `representation_plots.py` (the lightest ones) before moving to `fairness.py` and `summary_plot.py` where the semantic color logic lives. Keeping pixel-equivalent output throughout as you asked 🚀

Khanz9664 (commented):
@komoike-oss28-ui Thanks for addressing everything so quickly — I pulled the latest changes and the updates look good. The fallback warning and `n_bins` validation were exactly the sort of defensive hardening I was hoping for, and I appreciate the added regression coverage.

At this point I’m comfortable with the abstraction direction (`style.py`, scoped rcParams, semantic color separation, etc.), so I think we’ve cleared the architecture review stage.

That said, I’d still prefer we keep this PR open and continue the remaining migrations on the same branch before merging. Since #57 is specifically about *centralizing visualization styling*, I think we’ll get a much cleaner result if we land this once the visualization layer is consistently migrated rather than ending up in a half-centralized state.

I especially want to see how the abstraction holds up once `summary_plot.py` and `fairness.py` are migrated, since those contain the most semantic styling logic and will be the real stress test for avoiding drift while preserving visual parity.

So from my side: direction looks good, the latest fixes are solid, and please continue with the remaining modules on this PR.

komoike-oss28-ui:
Thanks @Khanz9664 🙏 Glad the abstraction is cleared.

I'll continue the remaining migrations on this branch in difficulty order: starting with `failure_plots.py` and `bias_plots.py` (lightest), then `representation_plots.py`, and saving `fairness.py` and `summary_plot.py` for last since those are the real stress tests for the semantic color logic — as you said. Visual parity stays pixel-equivalent throughout; any non-parity fix (like the `plt.show()` bug from last round) gets called out explicitly in the commit message.

Will push commits incrementally so you can sanity-check along the way if anything looks off. No need to review each one — I'll ping when I'm ready to flip to Ready for review 🚀

Khanz9664:
@komoike-oss28-ui Sounds good — that migration order makes sense to me. Saving `fairness.py` and especially `summary_plot.py` for last is probably the right call since they’ll be the real test of whether the semantic styling abstraction holds up cleanly.

Appreciate you being intentional about visual parity and calling out any deliberate non-parity fixes separately — that’ll make review much easier on my end.

Incremental commits are perfect. No need to wait for feedback between each migration unless you hit something architectural that feels awkward in the abstraction. Otherwise, feel free to keep moving and just ping me once the full migration pass is ready for review 🚀

komoike-oss28-ui:
Thanks @Khanz9664 🙏 Will keep moving through the remaining modules and ping once the full migration pass is ready for review 🚀

Khanz9664 (commented):
@komoike-oss28-ui Thanks for the update — I went through the representation migration and this looks very solid.

You’re following the migration plan exactly the way I was hoping for: keeping visual parity intact while progressively moving styling concerns into the centralized system. I especially appreciate the restraint around `tab10` in `plot_embedding_2d()` — leaving that untouched was the right call since it is effectively a separate color system and changing it would have introduced subtle visual drift.

The theme mappings also look clean (`green`/`amber`/`red` severity semantics, text colors, `savefig_dpi`, etc.), and I’m glad to see `plt.style.use(...)` removed in favor of `apply_style()` here.

Also thanks for restoring the dropped CHANGELOG entry and extending the migration list — good catch there.

No concerns from my side on this migration step. Please keep moving through the remaining modules as planned.

Visual parity staying effectively pixel-equivalent is still the right call for this PR 👍 Looking forward to the next migration pass.

komoike-oss28-ui:
Thanks @Khanz9664 🙏 Really glad the representation pass landed cleanly and that the `tab10` reasoning matched your thinking — that was the one spot where I paused, so good to have it confirmed.

Next up is `fairness.py`, which is the bigger of the two remaining stress tests (more functions, and `PALETTE` living alongside the violation colors). I'll map the `violation_colors` explicitly onto `theme.semantic["severity"]` and go function-by-function to keep parity verifiable, then `summary_plot.py` after that.

I'll keep pushing incrementally so you can sanity-check if anything looks off, and I'll ping when I'm ready to flip to Ready for review 🚀

komoike-oss28-ui:
Apologies for the delay in getting back to this — thanks for your patience 🙏

Update: all six plotting modules are now migrated to the centralized `style.py` 🎉

- `calibration_plots.py`, `failure_plots.py`, `bias_plots.py`, `representation_plots.py`, `fairness.py`, `summary_plot.py`

The two stress-test modules went as planned:
- **`fairness.py`** — `violation_colors` mapped to `theme.semantic["severity"]`; the unknown-severity fallback now resolves to `severity["unknown"]` (same `#CCCCCC`, just no longer a bare literal).
- **`summary_plot.py`** — this one previously called `plt.rcParams.update()` globally, leaking style into the user's session. It's now scoped inside `apply_style()`; I verified with a before/after `rcParams` diff that there's **no global leak** after rendering, which is exactly the goal of #57. Its grid-off / white-background look is preserved by re-applying only `axes.grid: False` inside the context (the base `seaborn-v0_8-whitegrid` enables the grid). Removing its private `_C` palette surfaced cross-module imports in `trust_score.py` / `report.py` (used in their `_repr_html_` output), so I repointed those to the canonical `BRAND_COLORS` — same hex values, identical rendered HTML.

I've also merged the latest `main` into the branch. The Deployment Summary UX from #115 added a few more `summary_plot._C` references in `report.py`, so I repointed those to `BRAND_COLORS` as well to keep the merged tree consistent.

Visual parity held throughout — full suite is green at 315 passed, no test edits. A few values are intentionally left as literals where there's no brand equivalent (the `#DDD` annotation-box edges, per-panel grid alphas); called out in the commit messages.

I've left the CHANGELOG as `(refs #57)` rather than `(closes #57)` for now — happy to flip it to `closes` whenever you're satisfied it's merge-ready, so the close lines up with the actual merge. Flipping to Ready for review now 🚀

Khanz9664 (commented):
@komoike-oss28-ui Thanks for the update and for carrying the migration through the remaining modules.

I reviewed the latest changes and the overall direction looks good. In particular, I appreciate the work to scope rcParams changes through `apply_style()` in `summary_plot.py` and the cleanup of duplicated palette definitions across the fairness visualizations.

Before merging, I'd like to do one final visual parity check around the categorical/brand color mappings (particularly the orange palette entries used in fairness and summary plots) to ensure they still resolve to the exact historical colors. Aside from that, the migration appears to satisfy the original goal of centralizing visualization styling while preserving behavior and avoiding global style leakage.

Thanks again for the thorough work on this.

komoike-oss28-ui:
Thanks @Khanz9664 — good call checking the oranges specifically, since `BRAND_COLORS` has both `orange` (#F5784B) and `deep_orange` (#FF6B35).

I diffed the pre-migration sources (at `8184fd7`, before fairness/summary were touched) against the centralized values. Every "orange" resolves to its exact historical hex:

- **fairness** — the `plot_equalized_odds` / `plot_fairness_gap` bars previously hardcoded `color="#F5784B"`; they now use `theme.brand["orange"]`, which is `#F5784B`. The inline `PALETTE` had `#F5784B` as its 2nd entry, and `get_categorical_colors()` returns the centralized `PALETTE` in the same order, so it resolves index-for-index identically.
- **summary** — the reliability gap fill and the class-distribution palette used `_C["orange"]` = `#F5784B`; they now use `BRAND_COLORS["orange"]` = `#F5784B`.
- **no cross-contamination** — `deep_orange` (#FF6B35) is only ever the 8th `PALETTE` entry, exactly as before. Nothing that used `#F5784B` now resolves to `#FF6B35` or vice versa.

So the categorical/brand orange mappings are bit-for-bit unchanged. Happy to walk through specific line ranges if it helps the final check. Thanks again for the careful review!

Khanz9664 (approved):
@komoike-oss28-ui Thanks for the detailed follow-up and for walking through the color mappings so thoroughly.

I completed the final review and independently verified the points discussed above, including the categorical palette mappings, semantic color usage, summary dashboard migration, HTML report updates, and the integration points introduced by recent changes on `main`.

The visual parity checks look good — the historical color assignments have been preserved, including the orange/deep_orange distinctions, and the centralized palette resolves correctly across the migrated modules. I also confirmed that the styling changes remain scoped through `apply_style()`, eliminating global matplotlib state leakage while preserving existing behavior and appearance.

At this point, the migration appears to fully satisfy the goals of #57:

* Centralized visualization styling
* Consolidated color definitions and semantic mappings
* Scoped matplotlib styling with no global leakage
* Consistent theme usage across visualization modules
* Backward-compatible behavior with preserved visual output

The implementation has been thoughtfully executed, the merge-related follow-up fixes were handled cleanly, and the test suite remains green.

Approved — thanks for the substantial contribution and for the care taken to preserve visual parity throughout the migration. 🚀

komoike-oss28-ui:
Thank you so much, @Khanz9664 — this really means a lot. 🙏

I appreciate the depth of your review: independently re-verifying the categorical and semantic mappings, the orange/deep_orange distinction, the summary dashboard migration, and especially the `main` integration points. That careful parity checking is exactly what gave me confidence the refactor was safe to land.

Thank you also for your patience through the merge-related follow-ups — pulling in the latest `main` surfaced a couple of `_C` references the new Deployment Summary work had introduced, and your steady review made them straightforward to resolve cleanly.

This was a genuinely rewarding issue to work on, and I learned a lot about the project's visualization layer along the way. If there's appetite for it down the line, I'd be glad to build on this foundation — e.g. a dark-mode theme or an accessibility-focused palette now that the styling is centralized. Thanks again for the thoughtful guidance and for maintaining such a welcoming project. 🚀

PR #120, opened 2026-06-03
fix: map xgboost Booster labels

### LINKED ISSUE #117: [BUG]: XGBoost raw Booster models fail to map string labels in multi-class prediction (causes downstream crash)
### 🟢 Problem
In `trustlens/backends/xgboost.py`, when a raw `xgb.Booster` is passed to the resolver instead of an `XGBClassifier`, it falls back to extracting `y_pred` via `np.argmax(y_prob, axis=1)`. Because a raw `Booster` does not have a `classes_` attribute, the predicted labels are always integer indices `[0, 1, 2...]`. 
If a user provides a `y_true` containing string labels (e.g., `["cat", "dog", "mouse"]`), the subsequent equality check `y_true == y_pred` in the core pipeline will silently fail (or evaluate to all `False`), causing massive metric distortion or type-casting crashes in downstream modules (like `misclassification_summary`).

### 🎯 Expected Outcome
The XGBoost backend should gracefully handle missing `classes_` by allowing users to inject a `classes` array via the `analyze` kwargs, or the core pipeline should safely map `y_true` string labels to ordinal integers before calling the backend resolver. 

### 🧩 Where to Edit
- `trustlens/backends/xgboost.py` (around Line 96 in `resolve()`)
- `trustlens/api.py` (to potentially pass down expected `class_labels`)

### ⚡ Proposed Steps
1. Add an optional `classes` argument to `PredictionBundle` or the `resolve` function signature.
2. If `model` is `xgb.Booster` and string labels are detected in `y_true` (which requires passing `y_true` or `classes` into the resolver), map the `argmax` indices to the corresponding string classes.
3. Write a test case in `tests/test_backends.py` passing an `xgb.Booster` with string target variables.

nanookclaw:
Fixes #117.

## Summary

This adds explicit class-label plumbing for prediction backends so raw `xgboost.Booster` multiclass outputs can map ordinal probability columns back to semantic labels. `analyze()` now accepts optional `class_labels` in probability-column order and passes that through the resolver and core pipeline. When a raw Booster falls back to `argmax(y_prob)`, it uses those labels when the label count matches the probability columns instead of returning bare integer indices.

The same label metadata is carried through `PredictionBundle`, manual overrides, and sklearn-compatible fallback paths. The multiclass Brier-score path now encodes semantic `y_true` labels against the probability column labels locally, while failure and class-imbalance summaries avoid forcing class keys through `int()`. That keeps the string-labelled analysis path consistent without changing numeric-label behavior.

## Verification

- `ruff check` on touched files
- `pytest tests/backends/test_xgboost_logic.py tests/test_backend_xgboost.py -q`
- `pytest -q`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * analyze accepts optional class_labels so users can supply custom string or non-numeric labels for multiclass evaluations.

* **Bug Fixes**
  * Better handling of non-numeric/NumPy-scalar class labels across prediction resolution, metrics, and misclassification reporting to avoid incorrect conversions.

* **Tests**
  * Added tests validating string-label mapping from probability outputs and validating error on mismatched class label counts.


Khanz9664 (changes requested):
@nanookclaw Thanks for working on this — I reviewed the implementation and the overall approach looks solid. The explicit `class_labels` plumbing through `analyze()`, the backend resolvers, and the core pipeline addresses the root cause described in #117, and the new regression test for raw `xgboost.Booster` + string labels is particularly valuable.

Before I can approve, I'd like to see one issue addressed:

In `_encode_labels_for_probability_columns()`, when `class_labels` is provided but its length does not match the number of probability columns, the code currently falls back to:

```python
return y_true.astype(int)
```

This appears to mask a contract violation and can either fail unexpectedly for string labels or produce incorrect encodings for non-ordinal labels. I think this should fail fast with a clear `ValueError` explaining that the supplied `class_labels` length does not match the probability matrix shape.

Additionally, there are currently failing Linting and Type Checking checks that should be resolved before merge.

The remaining comments (missing label detail in the error message, `_as_python_label` duplication, etc.) are reasonable follow-ups but I don't consider them blockers for this bug fix.

Once the contract validation issue and CI failures are addressed, I'll be happy to take another look.

nanookclaw:
Addressed this in commit `138a5b6`.

Changes:
- `_encode_labels_for_probability_columns()` now raises `ValueError` when supplied `class_labels` length does not match the probability column count.
- Added a focused regression test for the mismatch case.
- Applied the formatting fixes and resolved the mypy `no-any-return` issue.

Local verification passed:
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest tests/test_core_pipeline.py tests/backends/test_xgboost_logic.py tests/test_backend_xgboost.py -q` (12 passed)
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 pytest -q` (279 passed)
- `ruff format --check ...`
- `ruff check ...`
- `mypy trustlens/ --follow-imports=skip`
- `git diff --check`

Khanz9664 (approved):
@nanookclaw Thanks for the follow-up work on this.

I reviewed the latest changes, including the contract validation added in `138a5b6` and the subsequent merge from `main`. The mismatched `class_labels` case is now handled explicitly with a clear error path, the regression coverage is in place, and all CI checks are passing.

This resolves the concerns raised in my previous review. Thanks for the quick turnaround and for adding the focused test coverage.

Approved.

PR #142, opened 2026-06-18
feat: regression analysis path + TrustReport integration (#82)

Whatsonyourmind:
Phase 1 of the regression follow-up to #141, in the ordering you outlined (pipeline/report first; visualization and a regression trust score as later phases).

## What's in this PR (your Phase-1 scope)
- **Detect regression targets + route through a regression path** — `analyze(..., task="auto"|"classification"|"regression")`. Auto-detection deliberately errs toward classification (only clearly-continuous float targets route to regression, so a discrete label set is never mis-routed). The regression path skips the classification backend (no class probabilities) and resolves point predictions from `y_pred` or `model.predict(X)`.
- **Surface the regression metrics in the report object** — a new `_run_regression_pipeline` mirrors the classification pipeline (metrics → results → report) and routes through the merged metrics (`error_distribution`, `prediction_interval_coverage` / PICP, `error_variance_correlation`), under `report.results["regression"]`. The two uncertainty metrics degrade gracefully to `status="skipped"` when `prediction_intervals` / `predicted_variance` aren't supplied.
- **Regression report rendering** — `report.show()` renders a dedicated regression reliability summary (error distribution + PICP + error–uncertainty correlation, with a heavy-tail hint).
- **Serialization** — `report.to_dict()` and `report.save()` (json / txt / bundle) handle regression.

## Architecture / scope
Followed the existing classification architecture (metrics → report container → rendering) and kept the regression logic isolated. `TrustReport` gains a `task_type`; for regression the classification **Trust Score is intentionally not computed** (the separate phase you flagged). Classification-only features (`summary_plot`, `plot`, `plot_bias`, `show_failures`, `deployment_explanation`) raise a clear `NotImplementedError` on a regression report rather than failing cryptically — they'll light up with the visualization / trust-score phases. The classification path is unchanged.

## Usage
```python
# point predictions only
report = analyze(model, X_val, y_val, task="regression")
report.show()

# with uncertainty -> enables PICP + error–variance correlation
report = analyze(model, X_val, y_val,
                 prediction_intervals=(lower, upper),
                 predicted_variance=variance,
                 task="regression")
```

## Tests / checks
- `tests/test_regression_pipeline.py` (10): routing, auto-detection, `show()`, `to_dict()`, `save()` round-trip, the `model.predict` path, uncertainty metrics populated/skipped, and the classification-only guards.
- Full suite green (355 passed); `ruff` clean; `mypy trustlens/ --follow-imports=skip` clean.

[...]

PR #141, opened 2026-06-17
feat: add regression reliability metrics (EPE distribution, PICP, error-variance correlation)

Whatsonyourmind:
## What

Implements the regression-reliability metrics requested in #82 as a new `trustlens/metrics/regression.py` module (metrics-only — see Scope).

Three functions, matching the existing metric-module conventions (raw-array inputs, plain-dict returns, NumPy docstrings, graceful `status="skipped"` degradation):

- **`error_distribution(y_true, y_pred)`** — the EPE / absolute-error distribution: `median_absolute_error` (MedAE), `p90_absolute_error`, `max_error`, `mean_absolute_error`, `rmse`, plus histogram data for plotting.
- **`prediction_interval_coverage(y_true, lower, upper, confidence_level=0.95)`** — PICP. Returns coverage vs. nominal, `calibration_error`, `mean_interval_width`, and a verdict (over-/under-/well-calibrated). **Degrades gracefully** to `{"status": "skipped", ...}` when no intervals are supplied (sklearn regressors expose only point predictions).
- **`error_variance_correlation(y_true, y_pred, predicted_variance)`** — Pearson + Spearman between predicted uncertainty and actual `|error|` (does the model know where it's wrong?). Degrades gracefully when no variance is supplied; zero-variance inputs return `0.0` rather than `NaN`.

## Scope

Deliberately **metrics-module-only**, per `CONTRIBUTING.md` §5 ("`api.py` integration is optional Step 3") and the "one PR per feature" guideline. The issue's other DoD items — `analyze()` auto-dispatch on continuous targets, the visualization functions, and the `prediction_contract` / backend changes to admit regressors — are a larger cross-cutting change (they touch `api.py`, `core/pipeline.py`, `backends/sklearn.py`, `report.py`, `trust_score.py`) and are cleaner as follow-ups. This PR lands the correct, tested primitives; happy to follow up with the wire-in.

## Tests

`tests/test_regression.py` — **16 tests, all passing**: perfect-predictor and analytically-known MedAE/p90, `make_regression` integration (per the DoD), the graceful-skip paths, full/zero/normal-interval PICP, informative vs. uninformative variance correlation, constant-variance → 0, and shape-mismatch errors.

- `pytest tests/test_regression.py` → 16 passed
- existing `tests/test_metrics_all_exports.py` + `tests/test_api.py` still pass (no regressions)
- `ruff check`, `ruff format --check`, `mypy` all clean

Refs #82.

## Summary by CodeRabbit

* **New Features**
[...]

Issue #82, opened 2026-05-08
[FEATURE] Add Support for Regression Models (Uncertainty & Error Analysis)


Khanz9664:
#### **Problem**
TrustLens is currently optimized for classification tasks (providing Brier Scores, ECE, and Bias gaps). However, a significant portion of production ML involves **Regression** (predicting continuous values). Trusting a regression model requires more than just looking at $R^2$; it requires understanding **Uncertainty Calibration** and **Heteroscedasticity** (error variance).

#### **The Vision**
We want to enable `trustlens.analyze(model, X, y_true)` to automatically detect when a target is continuous and switch to the **Regression Pipeline**.

#### **Your Task**
Implement the following core metrics and visualizations in a new `trustlens/metrics/regression.py` module:

1.  **Expected Prediction Error (EPE) Distribution**:
    *   Instead of misclassification, report the distribution of absolute errors.
    *   **Metric**: Median Absolute Error (MedAE) and 90th percentile error.
2.  **Uncertainty Calibration (for Probabilistic Regression)**:
    *   If the model provides prediction intervals (e.g., Quantile Regression), implement **PICP (Prediction Interval Coverage Probability)**.
    *   A trustworthy model should have a coverage probability that matches the confidence level (e.g., 95% of points fall in the 95% interval).
3.  **Error-Confidence Correlation**:
    *   Measure the correlation between predicted variance (if available) and actual error.
4.  **Visualization**:
    *   `plot_residual_analysis()`: A plot showing residuals vs. predicted values to detect patterns (heteroscedasticity).
    *   `plot_error_distribution()`: A histogram of errors compared to a normal distribution.

#### **Expected API**
```python
import trustlens

# TrustLens should auto-detect regression based on y_true.dtype
report = trustlens.analyze(reg_model, X_val, y_val)

report.show()
# Outputs:
# Regression Analysis
# - Median Absolute Error: 1.24
# - Max Error: 15.3 (Potential Outlier Risk)
# - Uncertainty Coverage: 92% (Target: 95%) -> Warning: Under-confident
```

#### **Definition of Done**
- [ ] New `metrics/regression.py` with MedAE and Coverage metrics.
- [ ] New `visualization/regression_plots.py` with residual and error distribution plots.
- [ ] Logic in `api.py` to auto-dispatch to the regression pipeline if `y_true` is continuous.
- [ ] At least 3 unit tests using `sklearn.datasets.make_regression`.

---

## Completed
- [x] Regression metrics module
[...]

PR #143, opened 2026-06-18
feat(viz): regression residual + error-distribution plots (#82)

Whatsonyourmind:
## Regression visualizations (phase 3 of #82)

Follows the merged metrics (#141) and pipeline/report integration (#142). Implements the two visualizations from the #82 checklist, as agreed in #142.

### What's added

**`trustlens/visualization/regression_plots.py`**
- `plot_residuals(y_true, y_pred, prediction_intervals=None, …)` — residuals (`y_true − y_pred`) vs. predicted value: the canonical view for **heteroscedasticity** (a fan/curve in the cloud) and **bias** (the cloud drifting off zero). Includes a dashed zero-error reference, an optional **prediction-interval band** drawn in residual space (`lower − pred` to `upper − pred`), and a `bias` / `corr(|residual|, ŷ)` annotation.
- `plot_error_distribution(y_true, y_pred, …)` — signed-error histogram with a **fitted-normal overlay** (so skew / heavy tails are obvious) and an MAE/RMSE annotation.

Both follow the existing plot-module conventions: `apply_style()` theming, the `save_path` / `show` contract, and a returned `matplotlib.figure.Figure`. `matplotlib` stays lazily imported by the report (the module itself is imported inside the report methods, as the other viz modules are).

**`TrustReport` methods**
- `TrustReport.plot_residuals()` / `.plot_error_distribution()`, guarded by a new `_require_regression()` — the mirror of the existing `_require_classification()`; calling them on a classification report raises `NotImplementedError`. The classification guard message now points at these new plots.
- The report now retains `prediction_intervals` / `predicted_variance` (optional, `None` for classification) so the interval band is available end-to-end from `report.plot_residuals()`; wired through the regression pipeline. Backward-compatible (new constructor params default to `None`).

### Tests
`tests/test_regression_visualization.py` (16): figure/axes assertions, the interval band (present only when supplied), validation errors (shape mismatch, empty, `bins < 1`, mismatched bounds), `TrustReport` method integration, the classification guard, and a no-display (Agg) smoke test. Full suite **358 → 374**, all green; `ruff check`, `ruff format --check .`, and `mypy trustlens/ --follow-imports=skip` clean.

Scoped as a dedicated PR, separate from the regression trust score (which, as discussed, needs its own design pass on weighting/thresholds). Closing notes / screenshots happy to add if useful.

## Release Notes

* **New Features**
  * Added residuals and error-distribution visualization methods to regression reports.
  * Regression reports now optionally display prediction-interval uncertainty as bands on residuals.
  * Regression reports now store and surface prediction uncertainty details for later use.

* **Tests**
  * Added/updated headless regression-visualization tests to validate figure outputs, labels/legend entries, interval-band rendering, and input validation.
  * Added smoke coverage to ensure plots can be generated without display.

### DIFF
--- a/tests/test_regression_visualization.py
+++ b/tests/test_regression_visualization.py
@@ -55,11 +55,25 @@ class TestPlotResiduals:
         assert "Residuals" in _legend_labels(ax)
 
     def test_interval_band_added_when_supplied(self, reg_arrays):
+        from matplotlib.collections import PolyCollection
+
         _, y_true, y_pred = reg_arrays
         lo, hi = y_pred - 2.0, y_pred + 2.0
         fig = plot_residuals(y_true, y_pred, prediction_intervals=(lo, hi), show=False)
         ax = fig.axes[0]
         assert "Prediction interval" in _legend_labels(ax)
+        # The band must actually be drawn in residual space spanning ~[-2, +2]
+        # (lower - pred = -2, upper - pred = +2), not merely a legend entry.
+        bands = [c for c in ax.collections if isinstance(c, PolyCollection)]
+        assert bands, "expected a fill_between PolyCollection for the interval band"
+        ys = np.concatenate([p.vertices[:, 1] for p in bands[0].get_paths()])
+        assert ys.min() == pytest.approx(-2.0, abs=0.1)
+        assert ys.max() == pytest.approx(2.0, abs=0.1)
+
+    def test_malformed_intervals_raise(self, reg_arrays):
+        _, y_true, y_pred = reg_arrays
+        with pytest.raises(ValueError, match=r"\(lower, upper\)"):
+            plot_residuals(y_true, y_pred, prediction_intervals=(y_pred,), show=False)
 
     def test_no_band_without_intervals(self, reg_arrays):
         _, y_true, y_pred = reg_arrays
--- a/tests/test_regression_visualization.py
+++ b/tests/test_regression_visualization.py
@@ -1,0 +1,181 @@
+"""
+tests/test_regression_visualization.py.
+=======================================
+Tests for the regression visualization plotting functions and their TrustReport
+methods: residual analysis and error-distribution plots, the optional
+prediction-interval band, regression-only guards, and a no-display smoke test.
+"""
+
+from __future__ import annotations
+
+import matplotlib
+
+matplotlib.use("Agg")  # headless: no display required for the test run
+
+import matplotlib.pyplot as plt
+import numpy as np
+import pytest
+
+from trustlens import analyze
+from trustlens.report import TrustReport
+from trustlens.visualization.regression_plots import (
+    plot_error_distribution,
+    plot_residuals,
+)
+
+
+@pytest.fixture
+def reg_arrays():
+    rng = np.random.default_rng(0)
+    n = 200
+    X = rng.normal(size=(n, 3))
+    y_true = X @ np.array([1.5, -2.0, 0.7]) + rng.normal(scale=1.0, size=n)
+    y_pred = y_true + rng.normal(scale=1.0, size=n)
+    return X, y_true, y_pred
+
+
+def _legend_labels(ax) -> list[str]:
+    legend = ax.get_legend()
+    return [t.get_text() for t in legend.get_texts()] if legend is not None else []
+
+
+# --------------------------------------------------------------------------- #
+# Module-level functions
+# --------------------------------------------------------------------------- #
+class TestPlotResiduals:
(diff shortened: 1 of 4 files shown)


Whatsonyourmind:
Pushed `a4fda42` addressing the automated review:

**Fixed**
- **`prediction_intervals` arity (CodeRabbit)** — a malformed value (e.g. a 1-element tuple) now raises a clear `ValueError` instead of an `IndexError`. Added a test.
- **Band test only checked the legend (LlamaPReview P2)** — strengthened: it now asserts the `fill_between` `PolyCollection` is actually drawn, spanning the expected residual range (~[-2, +2]), not merely that a legend entry exists.

**Looked at but kept as-is, with reasoning**
- **`plt.close(fig)` before `return fig` (LlamaPReview P1, "returned figure unusable")** — verified this isn't the case here: `fig.savefig()` works fine on the returned figure after `plt.close()` (matplotlib retains the `Figure` and its artists; only pyplot's state-machine reference is dropped), confirmed even under warnings-as-errors. It's also the existing convention across the repo's plot modules (`calibration_plots.py` etc.), and the function's own `save_path` writes before closing — so I kept it consistent rather than diverge.
- **DRY on the save/show/close tail (P2)** — left to match the established per-module plot convention (`calibration_plots.py` / `bias_plots.py` repeat the same tail); happy to refactor that across all plot modules in a separate PR if you'd prefer.

Full suite 374 → 375; `ruff check`, `ruff format --check .`, and `mypy --follow-imports=skip` all clean.

Khanz9664 (approved):
@Whatsonyourmind

Thanks for the continued work on the regression support roadmap.

I've reviewed the implementation, tests, and follow-up changes. The scope is well aligned with the progression established by #141 and #142, and the visualization layer integrates cleanly with the existing TrustReport and visualization architecture.

* Regression-only visualization guards mirror the existing classification patterns cleanly.
* The prediction-interval overlay is implemented in residual space rather than raw prediction space, which makes the diagnostic substantially more useful.
* Input validation and shape handling are thorough.
* The visualization test coverage is strong, and I appreciate the follow-up enhancement to verify that the interval band is actually rendered rather than only checking legend entries.

I also reviewed the discussion around the `plt.close(fig)` pattern. Since this matches the existing visualization modules and the returned figure remains usable for downstream operations such as `savefig()`, I don't consider that a blocker for this PR.

Overall this is a solid addition that completes the visualization portion of the regression roadmap and keeps the implementation appropriately scoped.

Thanks again for the thoughtful incremental approach to the regression work. 🚀

Khanz9664:
@Whatsonyourmind

Thanks again for the continued contributions and for helping move the regression roadmap forward.

With the merge of #141 (Regression Metrics), #142 (Regression Pipeline & TrustReport Integration), and #143 (Regression Visualizations), the majority of the original scope from #82 is now complete:

✅ Regression metrics (Error Distribution, PICP, Error-Variance Correlation)

✅ Regression pipeline auto-dispatch

✅ Regression report rendering and serialization

✅ Residual analysis visualization

✅ Error distribution visualization

The main remaining item is the **Regression Trust Score**.

Unlike the previous pieces, this requires a bit more design work because the existing Trust Score framework was built around classification-oriented concepts (calibration, confidence behavior, deployment risk, etc.). For regression we need to determine:

* Which metrics should contribute to the score
* How those metrics should be weighted
* How uncertainty calibration (PICP) should influence trust
* Whether the scoring system should remain comparable to classification reports or be regression-specific
* How deployment verdicts should be generated for regression workloads

Rather than jumping directly into implementation, I'd recommend opening a design/RFC discussion first so we can align on the scoring philosophy before introducing a public API.

If you're interested in taking that on, feel free to open a proposal issue outlining a possible regression trust score framework and we can iterate on it together before implementation.

Thanks again for the thoughtful incremental approach throughout this feature series. 🚀

Whatsonyourmind:
Thanks for merging the series, and for the thoughtful breakdown of what the Regression Trust Score needs — agreed that the scoring philosophy should be settled before any public API.

I've opened #145 as a design/RFC proposal that addresses each of your five questions, grounded in the merged regression metrics (`error_distribution` / `prediction_interval_coverage` / `error_variance_correlation`) and the existing `trust_score.py`. The core proposal: keep the *interface* identical (0–100, A/B/C/D, `TrustScoreResult`, weight-redistribution, blockers) but use three regression-specific dimensions, and — the one point I'd most like your read on — weight uncertainty reliability **above** raw accuracy, on the logic that a low-R² model with honest intervals is safer to deploy than a high-R² model with decorative ones.

Happy to iterate there before I touch any implementation. 🚀

PR #146, opened 2026-06-19
Fix reliability curve quantile bin collapse

### LINKED ISSUE #144: [BUG] reliability_curve quantile bin collapse can produce invalid bin assignments
## Problem Statement

There appears to be a latent bug in `trustlens/metrics/calibration.py` within the `reliability_curve()` function when using `strategy="quantile"` on highly concentrated probability distributions.

Under certain degenerate conditions (for example, when all predicted probabilities are identical), quantile binning can collapse multiple requested bins into a single effective bin. This can ultimately produce invalid bin assignments and cause the function to raise a runtime exception.

Because `reliability_curve()` is a core calibration utility, this failure can propagate into calibration visualizations and any future metrics that depend on reliability-curve outputs.

---

## Expected Behavior

`reliability_curve()` should gracefully handle highly concentrated probability distributions, including cases where quantile binning collapses multiple requested bins into fewer effective bins.

The function should still return a valid reliability curve rather than raising an exception.

---

## Actual Behavior

For certain degenerate probability distributions, the function can fail with:

```text
ValueError: 'list' argument must have no negative elements
```

---

## Root Cause Analysis

The issue appears to stem from the interaction between quantile binning and duplicate bin-edge removal:

1. Many samples share the same predicted probability (for example, all predictions are `0.0`).
2. Quantile bin generation produces duplicate edges.
3. `np.unique(bin_edges)` collapses those duplicate edges.
4. In extreme cases, `bin_edges` can reduce to a single value (for example, `[0.0]`).
5. The slice `bin_edges[1:-1]` becomes an empty array.
6. `np.digitize(y_prob, [])` assigns all samples to index `0`.
7. The subsequent clipping step effectively becomes:

```python
np.clip(0, 0, -1)
```

because the effective number of bins is zero.

8. This results in invalid negative bin assignments.
9. A downstream `np.bincount()` call then raises:

```text
ValueError: 'list' argument must have no negative elements
```

---

## Steps to Reproduce

```python
import numpy as np

from trustlens.metrics.calibration import reliability_curve

y_true = np.zeros(100)
y_prob = np.zeros(100)

reliability_curve(
    y_true,
    y_prob,
    strategy="quantile",
)
```

---

## Impact

This bug can affect:

* Reliability diagrams
* Calibration visualizations
* Calibration debugging workflows
* Future calibration metrics that depend on `reliability_curve()`
* Degenerate or highly overconfident models
* Small datasets where quantile bins may collapse

Because the failure occurs inside a core calibration utility, it can prevent downstream reporting and visualization from executing successfully.

---

## Potential Fixes

Possible approaches include:

* Aligning the bin-assignment logic with the implementation used by `expected_calibration_error()`.
* Adding explicit handling for degenerate quantile-bin collapse scenarios.
* Guarding against zero effective bins before bin assignment and aggregation.
* Standardizing calibration binning behavior across:

  * `reliability_curve()`
  * `expected_calibration_error()`
  * future calibration metrics such as Maximum Calibration Error (MCE).

The exact implementation is left to the contributor.

---

## Suggested Tests

Add regression tests covering:

* `test_reliability_curve_quantile_zero_variance`
* `test_reliability_curve_quantile_duplicate_edges`
* `test_reliability_curve_quantile_bin_collapse`
* `test_reliability_curve_returns_valid_output_after_bin_collapse`

---

## Related Discussion

This issue was identified during review of:

* #134 — Maximum Calibration Error (MCE)

The discussion highlighted a broader calibration-stack consistency concern between `reliability_curve()` and `expected_calibration_error()` when quantile binning collapses duplicate boundaries.

---

## Credit

Huge thanks to @cschanhniem for identifying and diagnosing this edge case during discussion of the MCE feature request. 🚀

gaoflow:
## Summary
- handle collapsed quantile bin edges in `reliability_curve()` by returning a valid single-bin curve
- add regression coverage for zero-variance and duplicate-edge quantile predictions
- update the unreleased changelog

## Tests
- `.venv/bin/python -m pytest tests/test_calibration.py -q`
- `.venv/bin/python -m pytest tests/ -q`
- `.venv/bin/pre-commit run --files trustlens/metrics/calibration.py tests/test_calibration.py CHANGELOG.md`

Fixes #144

AI assistance was used under my direction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `reliability_curve(strategy="quantile")` to correctly handle cases where quantile bin edges collapse, including when all predicted probabilities are identical—now returns a valid single-bin curve instead of failing.
* **Tests**
  * Added unit tests covering zero-variance predictions and duplicated quantile edges for the quantile strategy.
* **Documentation**
  * Updated the unreleased changelog entry to reflect the reliability curve fix.


Khanz9664 (commented):
@gaoflow Thanks for the contribution and for jumping on #144 so quickly.

I've reviewed the implementation and overall this looks like a solid fix. The root cause described in the issue is addressed correctly by handling the collapsed quantile-bin case before reaching the `np.digitize` / `np.bincount` path, and the added regression tests provide good coverage for both zero-variance predictions and duplicate-edge scenarios.

One small suggestion before I merge:

In the new early-return block:

```python
if n_bins_actual == 0:
    return (
        np.array([float(y_true.mean())]),
        np.array([float(y_prob.mean())]),
        np.array([len(y_true)], dtype=int),
    )
```

would you mind switching the `.mean()` calls to `np.mean(...)`?

```python
if n_bins_actual == 0:
    return (
        np.array([float(np.mean(y_true))]),
        np.array([float(np.mean(y_prob))]),
        np.array([len(y_true)], dtype=int),
    )
```

This is a minor robustness improvement, but it ensures the early-return path behaves correctly for generic array-like inputs (e.g. Python lists) in addition to NumPy arrays, and keeps the implementation consistent with the rest of the calibration module's NumPy-based handling.

Other than that, the fix looks good, the tests are well targeted, and the changelog update is appreciated. Thanks again for the contribution 🚀

gaoflow:
Thanks for the review. I switched the collapsed-quantile early return to use `np.mean(y_true)` and `np.mean(y_prob)` as suggested, so the path also handles generic array-like inputs.

Local verification after the change:
- `python -m pytest tests/test_calibration.py -q` -> 21 passed
- `python -m pytest tests/ -q` -> 367 passed, 2 skipped
- `pre-commit run --files trustlens/metrics/calibration.py tests/test_calibration.py CHANGELOG.md` -> passed
- `git diff --check` -> passed

Khanz9664 (approved):
@gaoflow

Thanks for the quick follow-up.

I've reviewed the latest changes and confirmed that the robustness update suggested during review has been incorporated. Using `np.mean(...)` in the collapsed-quantile early-return path makes the implementation more consistent with the rest of the calibration module and ensures the edge-case handling works for generic array-like inputs as well as NumPy arrays.

The fix now:

* Resolves the crash described in #144.
* Handles fully-collapsed quantile bins safely.
* Returns a valid single-bin reliability curve for degenerate probability distributions.
* Includes targeted regression tests covering both zero-variance and duplicate-edge scenarios.
* Updates the changelog accordingly.

Thanks for the thorough validation and quick turnaround on the review feedback. This looks ready to go from my side. 🚀

PR #147, opened 2026-06-19
feat: regression Trust Score (regression_trust_score) — implements RFC #145

### LINKED ISSUE #82: [FEATURE] Add Support for Regression Models (Uncertainty & Error Analysis)

### LINKED ISSUE #145: RFC: Regression Trust Score — scoring framework proposal (follow-up to #82)
Thanks for merging #141, #142 and #143, and for laying out the open questions so clearly. As you suggested, I'm opening this as a design/RFC **before** any implementation so we can align on the scoring philosophy first. Everything below is a proposal for discussion — happy to iterate or be talked out of any of it.

It's grounded in what's already merged: the regression metrics in `trustlens/metrics/regression.py` (`error_distribution`, `prediction_interval_coverage`, `error_variance_correlation`) and the existing classification scorer in `trustlens/trust_score.py`.

## TL;DR
- Keep the **interface** identical to classification (`0–100`, A/B/C/D, `verdict`, `TrustScoreResult`, weight-redistribution for missing dimensions, hard blockers) so reports stay consistent and downstream code doesn't branch on task type.
- Make the **dimensions** regression-specific — propose three: **Accuracy/Skill**, **Interval Calibration (PICP)**, **Uncertainty Informativeness**.
- Propose a deliberately **reliability-weighted** philosophy (uncertainty ≥ raw accuracy). That's the main thing I'd like to align on.

## 1. Which metrics should contribute

| Dimension | Source (already merged) | Raw signal | → sub-score [0,100] |
|---|---|---|---|
| **Accuracy / Skill** | `error_distribution` + `Var(y_true)` | skill score `S = 1 − MSE/Var(y)` (i.e. R²) + tail ratio `p90/median` | `100 × clip(S, 0, 1)`, minus a heavy-tail penalty |
| **Interval Calibration** | `prediction_interval_coverage` | `calibration_error = picp − target` | `100 × (1 − clip(|calibration_error| / T, 0, 1))` |
| **Uncertainty Informativeness** | `error_variance_correlation` | `max(pearson, spearman)` | `100 × clip(corr, 0, 1)` |

Rationale:
- **Accuracy/Skill must be scale-free.** Raw MAE/RMSE can't be normalized to `[0,100]` without a reference and aren't comparable across datasets. A skill score against the naive "predict-the-mean" baseline (= R²) is the standard scale-free choice and is computable from data already in the report (`MSE = rmse²` from `error_distribution`, `Var(y)` from `y_true`). The `p90_absolute_error / median_absolute_error` ratio adds a heavy-tail signal that a mean-based R² hides.
- **Interval Calibration** is the regression analog of ECE — `prediction_interval_coverage` already returns `calibration_error` and a `verdict`; mapping `|calibration_error|` through a tolerance `T` mirrors how `_calibration_score` treats ECE today.
- **Uncertainty Informativeness** rewards "more uncertain exactly where more wrong" — `error_variance_correlation` already returns Pearson/Spearman + a `verdict`.

## 2. How they should be weighted — a philosophical proposal

The classification scorer weights calibration highest (`0.35`) because overconfidence drives deployment risk. I'd argue the regression analog is **stronger**: a low-R² model with *honest, well-calibrated* intervals is still safe to deploy (you can act on its stated uncertainty), whereas a high-R² model with *decorative* intervals is dangerous. So I'd propose weighting reliability ≥ raw accuracy:

```
uncertainty inputs present:
  accuracy 0.30, interval_calibration 0.40, uncertainty_informativeness 0.30
point predictions only (no intervals / variance):
  accuracy 1.00   (others redistributed away)
```

This is the key point I'd like your read on before anything else — it's a genuine inversion of the "accuracy is king" default, but I think it's the right call for a *trust* score. Worth disagreeing with.

Graceful degradation reuses the existing redistribution in `compute_trust_score`: a point-only regression report scores on Accuracy alone, exactly as a no-embeddings classification report drops Representation today.

## 3. How PICP should influence trust

Two ways, mirroring how ECE is both a sub-score and a penalty/blocker today:
1. As the **Interval Calibration** sub-score above.
2. As a **hard blocker**: a nominal-95% interval that actually covers ≤ ~80% is the regression equivalent of "confidently wrong" → force grade D regardless of accuracy, like the existing `ece_val > 0.1` blocker. Proposed default: `calibration_error < −0.10` (materially over-confident) → blocked.

## 4. Comparable to classification, or regression-specific?

Proposal: **comparable in interface, not in number.** Reuse `TrustScoreResult`, the 0–100 scale, the A/B/C/D thresholds and the verdict strings so the report UI and any downstream consumer don't special-case task type. But document explicitly that a regression `75` and a classification `75` are **not** the same currency (different underlying dimensions) — cross-task numeric comparison isn't meaningful and we shouldn't imply it. A `task_type` field on the result makes the distinction explicit and honest.

## 5. Deployment verdicts for regression

Reuse the blocker-hierarchy pattern from `compute_trust_score`. Proposed regression blockers (→ auto grade D), in priority order:
1. **Negative skill** — `S < 0` (worse than predicting the mean) → never deployable.
2. **Severe interval miscoverage** — `calibration_error < −0.10` (over-confident intervals).
3. **Decorative uncertainty** — intervals/variance are present *and consumed* but `max(pearson, spearman) ≈ 0` → the uncertainty signal is unsafe to gate on.

Below the blockers, the same `80 / 60 / 40` → A/B/C/D bands.

## Open questions back to you
- Comfortable with reliability-weighted defaults (§2), or do you prefer accuracy-led to stay intuitive?
- Skill baseline: predict-the-mean (R²) is simplest — do you also want to allow a user-supplied baseline (e.g. last-value for time series)?
- Should the tail ratio be a sub-score component or only a penalty?

If the direction looks right, I'm happy to follow up with a small PR: a `regression_trust_score` path in `trust_score.py` (dispatched by `task_type`, reusing `TrustScoreResult`) + tests, kept behind the same public interface — no API change until we've settled the philosophy here.

Whatsonyourmind:
Draft implementation of the regression Trust Score we converged on in #145. Opening as a **draft** so it's easy to review against the agreed design before it's marked ready — happy to iterate on any of the thresholds or the wiring.

This is the implementation half of #82 (the metrics landed in #141 / #142 / #143; this is the scorer on top of them).

## What it does

`regression_trust_score(results, y_true)` in `trustlens/trust_score.py` reuses the same `TrustScoreResult` interface as classification (0–100, A–D, `verdict`, weight redistribution), scoring three regression-native dimensions:

| Dimension | Default weight | Signal | Source metric |
|---|---|---|---|
| Accuracy / Skill | 0.30 | `S = 1 − MSE/Var(y)` (R² vs the mean-predictor), **minus** a `p90/median` heavy-tail dock on this dimension | `error_distribution` + `Var(y_true)` |
| Interval Calibration | 0.40 | `100 × (1 − clip(|calibration_error| / T, 0, 1))`, `T = 0.20` | `prediction_interval_coverage` |
| Uncertainty Informativeness | 0.30 | `100 × clip(max(pearson, spearman), 0, 1)` | `error_variance_correlation` |

- **Point-prediction-only reports** collapse to Accuracy = 1.00 via the existing redistribution — the two uncertainty dimensions drop out exactly as Representation does for a no-embeddings classification report.
- **Blockers → grade D:** negative skill (`R² < 0`); severe interval miscoverage (`calibration_error < −0.10`).
- **Penalties (not blockers):** a heavy tail docks the Accuracy/Skill dimension (capped at 50% of that sub-score); weak uncertainty correlation docks the composite — mirroring how ECE is both a calibration sub-score and a penalty in the classification scorer.
- **Comparability:** `TrustScoreResult` gains a `task_type` field. A regression `75` and a classification `75` share the interface but aggregate different dimensions, so they are not directly comparable — documented on the field, and `compare()` now refuses to rank a mixed-task batch rather than produce a silently-meaningless ranking.

## Integration

- `report.py`: regression reports now compute the score (previously deferred to `None`) and surface it through `show()`, `to_dict()` and `save()`. The classification-only narrative (`deployment_explanation`, `summary_plot`, patterns) stays guarded by `_require_classification()`.
- Exported as `trustlens.regression_trust_score`.

## Tests

- `tests/test_regression_trust_score.py` — one test per blocker, per penalty, weight redistribution (point-only / intervals-only / all-three / custom weights), the sub-score helpers, and the constant-target edge.
- Updated the two `test_regression_pipeline.py` assertions that encoded the old "score intentionally absent" behavior, plus a `compare()` mixed-task-type guard test.
- Local gate green: `pytest` (394 passed), `ruff check`, `ruff format --check`, `mypy trustlens/ --follow-imports=skip`.

## The obvious knobs

The exact constants — `T = 0.20` calibration tolerance, tail threshold 3.0 / scale 7.0 / 50% cap, weak-corr boundary 0.50 with a max-15 penalty, miscoverage −0.10 — all live in one block at the top of the regression section, so they're easy to retune if you'd weight any of them differently.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Regression reports now compute and expose a regression-specific Trust Score with letter grades (A–D) and numeric scores (0–100), including accuracy, interval calibration, and uncertainty informativeness sub-scores.
  * Regression report output and saved exports now include Trust Score, grade, and sub-score details.
  * Added/expanded regression metrics and visualizations for reliability, interval coverage, and uncertainty relationships.
* **Improvements**
  * Direct comparison now rejects mixed regression and classification batches, raising an error when task types differ.


Khanz9664:
@Whatsonyourmind

Thanks for opening this as a draft and for keeping the implementation closely aligned with the RFC discussion in #145.

I've started reviewing the design and implementation. Since this PR touches some of the most central pieces of TrustLens — `trust_score.py`, report serialization, score computation, cross-task behavior, and comparison semantics — I want to give it a more thorough review than a typical feature PR before we move it out of draft.

At a high level, I appreciate that you've:

* Kept the existing `TrustScoreResult` interface intact rather than introducing a parallel scoring system.
* Followed the blocker vs. penalty distinctions we converged on in the RFC.
* Reused the existing weight-redistribution philosophy instead of creating regression-specific exceptions.
* Added dedicated tests around scoring behavior, serialization, and mixed-task comparisons.

Over the next review pass I'll be looking closely at:

* The scoring philosophy and threshold choices (weights, penalties, and blocker boundaries).
* Whether the regression score remains internally consistent with the broader TrustLens trust framework.
* Cross-task behavior and report comparison semantics.
* Serialization/export behavior and backward compatibility implications.
* Edge cases around missing uncertainty information and degenerate targets.

Nothing immediately stands out as fundamentally conflicting with the RFC, but given that this will become part of TrustLens' core scoring API, I want to spend some time validating both the implementation details and the long-term design implications before we move toward final approval.

Thanks again for taking the RFC-first approach here — it made reviewing the implementation significantly easier. I'll continue working through the code and leave more specific feedback as I go.

Khanz9664 (approved):
@Whatsonyourmind

I've completed a deeper review pass on this implementation and compared it against both the agreed RFC in #145 and the surrounding TrustLens architecture.

Overall, this is a very strong piece of work and one of the more significant additions to the project so far. The implementation stays remarkably faithful to the design discussion while avoiding the temptation to introduce a parallel scoring framework or task-specific APIs.

A few things that stood out positively during review:

* Reusing `TrustScoreResult` instead of introducing a separate regression-specific result type was the right architectural decision. It preserves API consistency for downstream consumers while still allowing regression-specific behavior through `task_type`.
* The addition of `task_type` and the corresponding guard in `compare()` is a particularly good safety mechanism. Regression and classification scores now share an interface without creating the false impression that they are directly comparable quantities.
* The blocker vs. penalty separation matches the RFC discussion closely. Negative skill and severe interval miscoverage are treated as hard failures, while heavy-tail behavior and weak uncertainty correlation appropriately influence trust without automatically forcing deployment rejection.
* Weight redistribution is implemented cleanly and preserves the existing TrustLens philosophy. Point-prediction-only regressors degrade gracefully onto the Accuracy/Skill dimension without special-case logic.
* The regression-specific scoring helpers are well factored and make the implementation substantially easier to reason about and maintain.
* Test coverage is comprehensive and covers the most important behavioral guarantees: blockers, penalties, redistribution, serialization, task tagging, and mixed-task comparisons.

I also reviewed the automated feedback that was raised during the PR:

* The TXT serialization concern appears to be a false positive. From the current implementation path, `_show_regression()` now surfaces the Trust Score and the existing text-generation flow should inherit that output automatically. I don't currently see evidence of a missing-score bug there.
* The constant-target edge case is mathematically sound as implemented. For a constant target, a perfect fit receives full skill while an imperfect fit collapses to zero skill. That behavior is reasonable and defensible. However, I do think the additional imperfect-constant-target test suggested by the review bot would be valuable as a regression test, not because I believe there is a bug, but because it locks down the intended behavior for a somewhat unusual edge case.

One observation that may be worth tracking separately as a future enhancement:

* The regression scorer currently depends on `y_true` to compute `Var(y)` and therefore cannot be recomputed from serialized metric results alone. I don't view that as a blocker for this PR, but it may be worth considering whether target variance should eventually be persisted as part of the regression metrics payload so that Trust Scores can be reproduced entirely from stored results.

I also read through the discussion raised by @DanceNitra regarding finite-sample instability in the blocker thresholds. I agree with the conclusion reached in the RFC thread: the concern is legitimate, but confidence-aware blockers, Wilson intervals, minimum-sample requirements, and significance-aware gating feel more appropriate as a follow-up enhancement once we have real-world usage data for the v1 thresholds. I would not delay this implementation for that additional complexity.

At this point I don't see any architectural issues, API concerns, or RFC deviations that would prevent this from moving forward. The implementation is consistent with the design we agreed on in #145, the safety mechanisms are sensible, and the resulting API remains coherent with the broader TrustLens trust-scoring framework.

Nice work on both the design process and the implementation. 🚀

Whatsonyourmind:
Thank you — both for the depth of the review and for the architectural read-back; it's genuinely useful to see the design intent reflected so precisely.

I've folded in the one concrete item from your review and marked the PR ready:

- **Imperfect-constant-target regression test** (added): locks down the edge exactly as you described — with `Var(y)=0`, a perfect fit earns full skill while an imperfect fit collapses to *zero* skill. The test also asserts the part that's easy to get wrong: because that skill is `0.0` (not negative), it is **not** the negative-skill blocker — it degrades gracefully onto a zero Accuracy/Skill dimension, and point-only redistribution carries that onto the composite (grade D, `is_blocked=False`). Covers both the helper (`_regression_accuracy_score(..., 0.0)["skill"] == 0.0`) and the end-to-end result.
- **TXT serialization** — agreed, false positive. `_show_regression()` surfaces the score and the text flow inherits it; no missing-score path.

On the **`Var(y)` reproducibility** point — fully agree, and I think it's the right call to keep it out of this PR. Persisting target variance into the regression metrics payload so a Trust Score can be recomputed from stored results alone is a clean, self-contained follow-up. Happy to open a tracking issue for it (and take the implementation) if you'd like it on the board — just say the word and I'll write it up referencing this thread.

I also rebased onto the current `main` (the merge you pushed in, plus #146/#99) — no conflicts, and the full suite is green locally (398 passed, ruff/format/mypy clean). Marked ready for review.

Thanks again for the RFC-first process on this one; designing it in #145 before writing code made the whole thing much cleaner to land.

Khanz9664 (approved):
@Whatsonyourmind

Thanks for following up on this.

I've reviewed the additional commit and the new constant-target test coverage, and this addresses the remaining edge case I wanted to see locked down before merge.

The new test does a good job of capturing the distinction between:

* **Negative skill** (`R² < 0`) → blocker
* **Zero skill** (`R² = 0` equivalent behavior in the constant-target edge case) → not blocked, but naturally receives a low trust score

That's an important behavioral boundary for the regression scorer, and having both the helper-level assertion and the end-to-end `regression_trust_score()` verification makes the intended semantics very clear.

I also rechecked the broader implementation against the RFC discussion and the architectural goals we agreed on:

* TrustScoreResult reuse remains clean and consistent.
* Regression/classification comparability is handled safely through `task_type` and comparison guards.
* Weight redistribution behaves as expected when uncertainty signals are absent.
* Blockers and penalties align with the finalized RFC.
* The additional test strengthens confidence around one of the trickier regression edge cases without adding complexity to the scoring logic itself.

One small item before merge: could you please update the CHANGELOG as well?

The original regression metrics PR (#141) included a changelog entry, but the subsequent regression feature PRs (#142, #143), and this trust score implementation (#147), don't appear to be reflected under the current Unreleased section. Since this PR effectively completes the regression trust scoring portion of the feature series, I'd like to make sure the changelog captures the major additions that users will see in the next release.

A concise entry describing the regression trust score, grading system, deployment verdicts, and regression report integration would be sufficient.

Other than that, I don't see any remaining technical concerns that would block merging.

Thanks again for taking the RFC-first approach on this feature series and for being responsive throughout review. This has been a thoughtful implementation from design through testing and it provides a strong foundation for regression trustworthiness analysis in TrustLens.

Khanz9664 (approved):
@Whatsonyourmind  I've added the missing changelog entries for the regression pipeline, visualizations, and trust score work so the regression feature series is fully represented in the upcoming release notes.

Everything else looks good. Thanks again for the thoughtful implementation and iteration throughout the RFC and review process. 🚀

Whatsonyourmind:
Thanks for merging this, @Khanz9664 — glad the regression Trust Score landed cleanly. Designing it in the #145 RFC first really did pay off: the implementation tracked the agreed v1 spec (3 dimensions, blocker-vs-penalty split, reliability-weighted defaults, mean-predictor baseline) without surprises, and the regression path now completes the #82 scope end to end.

One open item from the review thread, whenever it's useful (no rush): the `Var(y)` persistence follow-up — storing the target variance into the regression-metrics payload so a Trust Score can be recomputed from stored results alone. Happy to open a tracking issue and take the implementation if you'd like it on the board; otherwise I'll leave it parked.

Thanks again for the thoughtful review process throughout — it was a genuinely good collaboration.

Khanz9664:
@Whatsonyourmind

Thanks for the thoughtful collaboration throughout this entire regression feature series.

I agree that the RFC-first approach made a significant difference here. By the time we reached implementation, most of the important design decisions had already been discussed and documented, which made review substantially easier and helped keep the implementation aligned with the intended TrustLens philosophy.

On the `Var(y)` persistence follow-up: I do think that's a worthwhile enhancement, but I agree it belongs as a separate piece of work rather than something that should have been folded into #147.

Being able to recompute regression trust scores from stored analysis artifacts without requiring access to the original target array would improve reproducibility and make exported results more self-contained. It also aligns nicely with the broader goal of making TrustLens reports portable across environments.

Feel free to open a tracking issue outlining the proposed approach, affected components, and any API considerations you foresee. Once it's written up, I'd be happy to review the design and we can determine the best implementation path from there.

Thanks again for all the work across #141, #142, #143, #145, and #147. The regression support that exists today is substantially stronger because of that effort. 🚀

PR #151, opened 2026-06-24
feat(regression): persist Var(y) to recompute Trust Score from stored reports (#150)

### LINKED ISSUE #150: Persist Var(y) (target variance) in regression metrics payload to recompute regression Trust Score from stored reports
### Summary

Persist the target variance `Var(y)` into the regression metrics payload so that `regression_trust_score(...)` can recompute a regression Trust Score from a **stored** report, without needing the original `y_true` array in memory.

Follow-up to the regression Trust Score work (RFC #145, implemented in #147). Filing this as the tracking issue per the maintainer's suggestion on #147.

### Motivation

`regression_trust_score(results, y_true)` derives the skill score's denominator on the fly:

```python
y_true_arr = np.asarray(y_true, dtype=float)
target_variance = float(np.var(y_true_arr)) if y_true_arr.size else 0.0
...
accuracy = _regression_accuracy_score(error_dist, target_variance)  # S = 1 − MSE/Var(y)
```

So the score is only reproducible while the caller still holds `y_true`. Once a run is serialized to disk (or shipped to a dashboard / CI artifact / audit log) and the raw targets are gone, the regression Trust Score **cannot be recomputed** — even though every other input it needs (`error_distribution`, `interval_coverage`, `error_variance_correlation`) already lives in `results`. `Var(y)` is the one missing scalar.

This matters for the natural use cases of a Trust Score: re-grade a saved analysis after a weight-policy change, diff trust over time from stored reports, or verify a reported grade in review — all of which should work from the persisted artifact alone.

### Proposed approach

Persist `Var(y)` as an optional, backward-compatible field on the regression metrics payload at analysis time, and let `regression_trust_score` use it as a fallback when `y_true` isn't supplied.

1. **Write path (regression pipeline):** when the regression metrics are built, record the target variance alongside the existing keys, e.g. `results["regression"]["error_distribution"]["target_variance"]` (or a dedicated `results["regression"]["target_variance"]` — happy to follow whichever placement you prefer).
2. **Read path (`regression_trust_score`):** make `y_true` optional and resolve `target_variance` in priority order:
   - if `y_true` is provided → compute `np.var(y_true)` as today (no behavior change);
   - else if a stored `target_variance` is present in `results` → use it;
   - else → raise a clear `ValueError` explaining that one of the two is required (rather than silently producing `target_variance=0.0`, which currently routes through the "constant target" branch in `_regression_accuracy_score` and would yield a misleading skill score).

### Affected components

- `trustlens/trust_score.py`
  - `regression_trust_score` — make `y_true` optional; add the stored-variance fallback + the explicit error.
  - `_regression_accuracy_score` — unchanged (still takes a `target_variance: float`), but worth a doc note that `0.0` now means a genuine constant target, never "missing".
- The regression pipeline that assembles `results["regression"]` — emit `target_variance` on the metrics payload.
- Serialization of `TrustScoreResult` / the regression report (JSON + the human-readable TXT) — make sure the new field round-trips so a reloaded report is self-sufficient.

### API & compatibility considerations

- **Backward compatible:** the new payload field is optional and additive; older stored reports (no `target_variance`) keep working as long as `y_true` is passed, and the call signature stays source-compatible because `y_true` only becomes *optional*, not removed.
- **Schema/versioning:** if the report payload carries a schema version, bump it; otherwise the presence/absence of the field is itself a safe sentinel.
- **Numerical consistency:** persist the same `float(np.var(...))` (population variance, `ddof=0`) used today so a recompute-from-stored score is byte-for-byte identical to the in-memory one.
- **Edge case:** define behavior when both `y_true` and a stored `target_variance` are present and disagree — I'd suggest preferring the explicitly-passed `y_true` and (optionally) warning on a mismatch beyond a tolerance.

### Scope

Happy to implement this end-to-end (write path + read-path fallback + serialization round-trip + tests covering: recompute-from-stored equals recompute-from-`y_true`, missing-both raises, and a constant-target `Var(y)=0` case) and open a PR against the placement you prefer. Just confirm whether you'd like `target_variance` nested under `error_distribution` or as a top-level `regression` key.

Refs: #145 (RFC), #147 (regression Trust Score implementation).

Whatsonyourmind:
Closes #150. Follow-up to the regression Trust Score (RFC #145, implemented in #147).

Persists the target variance so `regression_trust_score(...)` can be recomputed from a **stored** report (JSON artifact, dashboard, CI artifact, audit log) without the original `y_true` in memory — `Var(y)` was the one missing scalar.

### What changed (per the placement/decisions you confirmed on #150)

- **Write path** (`core/pipeline.py`): emit a top-level `results["regression"]["target_variance"] = float(np.var(np.asarray(y_true)))` — population variance, `ddof=0`, byte-identical to the on-the-fly computation, so a recompute-from-stored score equals the in-memory one.
- **Read path** (`trust_score.py::regression_trust_score`): `y_true` is now **optional**; `Var(y)` resolves in priority order
  1. explicit `y_true` → `np.var` (no behaviour change),
  2. else persisted `target_variance`,
  3. else **`ValueError`** — never a silent `0.0` (which would route through the genuine constant-target branch and yield a misleading skill score).
  When both are present, the explicit `y_true` wins and a mismatch beyond tolerance emits a `UserWarning`.
- `_regression_accuracy_score`: doc note that `0.0` now unambiguously means a genuine constant target (missing variance is a `ValueError` upstream, never `0.0`).

### Compatibility
- Additive, optional field; older stored reports keep working as long as `y_true` is passed. The signature stays source-compatible (`y_true` only becomes optional). Serialization is unchanged — the float rides the existing `_to_serializable` path and round-trips in JSON automatically.

### Tests (all green locally)
`tests/test_regression_trust_score.py` adds: recompute-from-stored equals recompute-from-`y_true` (score/grade/sub-scores/breakdown); missing-both raises `ValueError`; a stored `Var(y)=0` constant target is honoured (not treated as missing); mismatch warns and `y_true` wins; a matching stored value does not warn. Full local run: `test_regression_trust_score`, `test_regression_pipeline`, `test_core_pipeline`, `test_regression`, `test_api`, `test_output_formatting`, `test_metrics_all_exports` — all pass (94 tests). End-to-end smoke confirms the field is emitted, survives a JSON round-trip, and recomputes to the same score with no `y_true`.

Happy to adjust placement/naming or the mismatch policy if you'd prefer a different convention.

## Summary by CodeRabbit

* **New Features**
  * Regression Trust Score can now be computed without providing original target values when the report already includes stored target variance.
  * Regression results now carry stored target-variance metadata for reuse.
* **Bug Fixes**
  * Improved handling for missing/empty target values, avoiding invalid stored values.
  * Clear error is raised when neither original targets nor stored variance is available.
  * If provided targets conflict with stored variance, a warning is emitted (and the provided targets are used); no warning when they match.
* **Tests**
  * Added regression coverage for persisted variance recomputation behavior.

### DIFF
--- a/tests/test_regression_trust_score.py
+++ b/tests/test_regression_trust_score.py
@@ -16,6 +16,8 @@ Covers the converged v1 spec:
 
 from __future__ import annotations
 
+import warnings
+
 import numpy as np
 import pytest
 
@@ -322,3 +324,67 @@ def test_constant_target_imperfect_fit_zero_skill_not_blocked():
     assert not r.is_blocked  # skill 0.0 is not < 0 → not a blocker
     assert r.sub_scores["accuracy"] == pytest.approx(0.0)
     assert r.grade == "D"
+
+
+# ---------------------------------------------------------------------------
+# Persisted target_variance — recompute a regression Trust Score from a stored
+# report without the original y_true (issue #150)
+# ---------------------------------------------------------------------------
+
+
+def test_stored_variance_recompute_equals_y_true():
+    """A score recomputed from a persisted target_variance is identical to the
+    score computed from y_true."""
+    res = _results(_ed(0.9), _cov(0.0), _corr(0.9))
+    from_y_true = regression_trust_score(res, Y)
+    res["regression"]["target_variance"] = VAR_Y  # what the pipeline persists
+    from_stored = regression_trust_score(res)  # no y_true in memory
+    assert from_stored.score == from_y_true.score
+    assert from_stored.grade == from_y_true.grade
+    assert from_stored.sub_scores == from_y_true.sub_scores
+    assert from_stored.breakdown == from_y_true.breakdown
+
+
+def test_missing_both_y_true_and_stored_variance_raises():
+    """Neither y_true nor a persisted variance → explicit error, never a silent
+    0.0 that would masquerade as a constant target."""
+    res = _results(_ed(0.9))  # no target_variance persisted
+    with pytest.raises(ValueError, match="target variance"):
+        regression_trust_score(res)  # and no y_true
+
+
+def test_stored_zero_variance_is_honoured_not_treated_as_missing():
+    """A genuine constant target persists as target_variance == 0.0 and must be
+    used (constant-target branch), not rejected as 'missing'."""
+    ed = _ed(0.0)
+    ed["rmse"] = 0.0  # perfect fit on a constant target
+    res = {
+        "regression": {
+            "error_distribution": ed,
+            "interval_coverage": _cov_skipped(),
+            "error_variance_correlation": _corr_skipped(),
+            "target_variance": 0.0,
+        }
+    }
+    r = regression_trust_score(res)  # no y_true; 0.0 must be used, not rejected
+    assert r.task_type == "regression"
+    assert r.sub_scores["accuracy"] == pytest.approx(100.0)  # perfect constant-target fit
+
+
+def test_y_true_wins_and_warns_on_mismatch_with_stored():
+    """When both are present and disagree, y_true wins and a warning fires."""
+    res = _results(_ed(0.9), _cov(0.0), _corr(0.9))
+    res["regression"]["target_variance"] = VAR_Y * 4.0  # deliberately wrong
+    with pytest.warns(UserWarning, match="disagrees with the persisted"):
+        mismatched = regression_trust_score(res, Y)
+    clean = regression_trust_score(_results(_ed(0.9), _cov(0.0), _corr(0.9)), Y)
+    assert mismatched.score == clean.score  # y_true wins → stored value ignored
+
+
+def test_matching_stored_variance_does_not_warn():
+    """A persisted variance equal to Var(y_true) must not warn."""
+    res = _results(_ed(0.9), _cov(0.0), _corr(0.9))
+    res["regression"]["target_variance"] = VAR_Y
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")  # any warning becomes a test failure
+        regression_trust_score(res, Y)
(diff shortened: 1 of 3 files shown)


Whatsonyourmind:
Thanks to the automated reviewers for the pass. Addressed the one actionable item in 994f8ad:

**Empty `y_true` → NaN persistence (LlamaPReview P1).** `np.var([])` returns `NaN`, which isn't JSON-serializable and would silently corrupt a recompute-from-stored score. The write path now falls back to `0.0` for an empty target array, which is exactly the value the read path already assigns an empty `y_true` (routing through the constant-target branch), so write and read stay consistent. Full regression + pipeline suite stays green (76 passed locally).

CodeRabbit reported no actionable items. Happy to adjust naming or the mismatch-warning policy if you'd prefer a different convention before merge.

Khanz9664 (approved):
@Whatsonyourmind

I've completed a review of the implementation and compared it against both the design discussion in #150 and the behavior we agreed on during the #147 follow-up conversation.

Overall, this looks good to me.

A few things I particularly appreciate:

* The placement of `target_variance` as a top-level regression field keeps the schema semantically clean. It represents a property of the target distribution rather than an error metric, so I think this is the right location.
* The resolution order is exactly what we discussed: explicit `y_true` → persisted `target_variance` → clear `ValueError`. Avoiding a silent fallback to `0.0` prevents missing data from being mistaken for a genuine constant-target case.
* The mismatch handling strikes a good balance between safety and usability. Using `y_true` as the source of truth while emitting a warning makes discrepancies visible without creating unnecessary friction.
* The empty-array guard added in the follow-up commit was an important catch. Keeping the persisted value consistent with the existing read-path behavior avoids NaN propagation and preserves reproducibility.
* The test coverage is thorough and exercises the key scenarios I was most interested in seeing: recompute-from-stored parity, missing-both failure, genuine zero-variance targets, warning behavior, and non-warning behavior.

Most importantly, this closes the reproducibility gap that motivated #150 without changing the scoring philosophy or behavior introduced in #147. Existing workflows remain compatible, while stored regression reports become substantially more self-contained and portable.

At this point I don't see any remaining technical concerns that would block merging.

Thanks again for following through on the follow-up and for keeping the implementation closely aligned with the design discussion.

Issue #155, opened 2026-06-27
RFC: Interval-calibration extensions — per-quantile coverage + CRPS (follow-up to #145 / v0.5.0)


Whatsonyourmind:
Follow-up to the v0.5.0 regression Trust Score (#145). Framing these as **extensions to the existing calibration / informativeness dimensions**, not new top-level concepts, per the direction in the v0.5.0 release thread.

## TL;DR

The current Interval Calibration dimension keys on a single prediction-interval level (PICP at one nominal level, e.g. 90%). That under-determines calibration: a model can be well-calibrated at 90% and badly miscalibrated everywhere else. Two extensions deepen the existing dimensions without adding new concepts, and both **degrade gracefully** to today's behavior when richer inputs aren't available:

1. **Per-quantile coverage → Interval Calibration.** Evaluate empirical coverage against nominal across several levels, summarized as one Interval Calibration Error (ICE) scalar + a reliability table for the report.
2. **CRPS → Uncertainty Informativeness.** A proper scoring rule that rewards sharp *and* calibrated predictive distributions in a single number, with an optional skill-score form (CRPSS) against a climatology reference.

## 1. The gap: single-level PICP under-determines calibration

PICP at one level answers "is the 90% PI's coverage ≈ 90%?" It can't see whether the 50% or 80% intervals are calibrated. Two models with identical PICP@90 can have very different calibration across the distribution. The fix is to sample coverage at multiple nominal levels — a discretized reliability curve for regression intervals.

## 2. Per-quantile coverage (Interval Calibration extension)

For nominal levels τ ∈ {0.1, 0.2, …, 0.9} (configurable), compute empirical central-interval coverage `emp(τ)` and define:

```
ICE = mean_τ | emp(τ) − τ |        # interval calibration error, lower is better
```

- **Inputs / graceful degradation:** needs quantile predictions at multiple levels, or a predictive distribution to derive them. If only a single PI is available, fall back to today's PICP (no behavior change). If only point predictions, the dimension stays inactive (existing weight-redistribution handles it).
- **Report surface:** one ICE scalar feeding the calibration dimension + a reliability table (nominal vs empirical) reusing the v0.5.0 visualization infra.
- **Interpretation:** ICE is a regression analogue of ECE; it makes "calibrated" mean "across the distribution," not "at one level."

## 3. CRPS (Uncertainty Informativeness extension)

[...]

PR #161, opened 2026-06-29
feat(regression): interval-calibration extensions — ICE + calibration-conditioned sharpness (RFC #155)

Whatsonyourmind:
Implements the design converged in #155 (following the RFC-first cadence from #145).

### What this adds
- **`multilevel_interval_coverage()`** (`metrics/regression.py`) — evaluates a set of prediction-interval levels at once and summarises them with:
  - **ICE** (Interval Calibration Error): `mean_tau |emp(tau) - tau|`, the multi-level analog of `|PICP calibration_error|`.
  - **sharpness_skill**: a calibration-conditioned sharpness proxy. Among only the levels that pass calibration (`|emp(tau) - tau| <= tol`), `1 - mean(model_width / climatology_width)`. Over-confident-narrow intervals fail the gate and are excluded, so they can't inflate it — the "tightness only where honest" property (CRPS-Resolution in spirit).
- **Trust Score wiring** (`trust_score.py`): Interval Calibration scores from ICE when multi-level data is present (single-level PICP fallback); Uncertainty Informativeness prefers the sharpness proxy (error-variance correlation fallback). The severe-miscoverage blocker keys on the worst per-level gap; the weak-correlation penalty only applies on the correlation path (no double-count).
- **End-to-end** (`api.py`, `core/pipeline.py`, `report.py`): `prediction_intervals` accepts either the existing `(lower, upper)` tuple or a `{level: (lower, upper)}` mapping. `report.show()` renders a multi-level block (ICE / sharpness skill / worst gap); data persists via the results dict.

### Design notes (per the RFC)
- **Three pillars preserved**, each a distinct property: Accuracy = central estimate, Interval Calibration = reliability, Informativeness = resolution.
- **ICE is a penalty signal, not a blocker** — same call as moving uncertainty correlation to a penalty in #145; we don't yet have empirical ICE distributions to justify a hard Grade-D threshold.
- **Proxy-first**: the calibration-conditioned sharpness proxy reuses the per-level coverage machinery (no separate estimator). It measures the same property as the full CRPS Reliability/Resolution decomposition, so swapping that in later is drop-in.
- **Raw CRPS/CRPSS deferred to a report diagnostic** — kept out of the weighted score so calibration isn't counted twice.

### Compatibility & tests
- Fully backward compatible: single-level / point-only / variance-only reports score exactly as before.
- New tests: metric unit tests (incl. the over-confident-narrow level being excluded from sharpness), trust-score integration (ICE + proxy paths, fallbacks, blocker), pipeline integration (dict intervals → ICE → report). Full suite green (424 passed).

Happy to adjust naming, tolerances, or the report layout to your taste.

## Summary by CodeRabbit

* **New Features**
  * Added multi-level prediction-interval calibration for regression, including ICE, calibrated-level counts, and a calibration-conditioned sharpness proxy.
  * Extended `analyze()` to accept `{confidence_level: (lower, upper)}` prediction-interval mappings.
  * Regression Trust Score now includes `informativeness_status` to explain uncertainty informativeness outcomes.
* **Changed**
  * Regression Trust Score and console output prioritize multi-level “Interval Calibration (multi-level)” when available.
* **Bug Fixes**
  * Improved validation and maintained backward-compatible single-level reporting.
* **Tests**
  * Added regression and Trust Score RFC-style test coverage for routing, rendering, and edge cases.


Whatsonyourmind:
Thanks @coderabbitai and @llamapreview — went through the findings. Pushed da8c15e addressing the substantive ones:

- **Sharpness from raw widths** (CodeRabbit): `sharpness_skill` now divides raw model widths by raw climatology widths instead of the 4dp-rounded `per_level` values, so display rounding can't bias it. Climatology widths are also computed once, vectorized over levels (addresses LlamaPReview's per-level `np.quantile` note).
- **Stronger sharpness test** (CodeRabbit): the positive-skill test previously used intervals drawn from the same marginal law, so it only passed by sampling noise. It now centers intervals on the conditional mean (spanning just the noise) over a signal-plus-noise target — genuinely tighter than the marginal spread while staying calibrated, so the assertion is robust.
- **Representative band** (CodeRabbit): the legacy single-interval/plot field now picks the widest band by mean width rather than assuming the highest nominal level is outermost (correct even if levels cross).
- **`confidence_level` docs** (LlamaPReview): clarified it applies only to the single-tuple path; in the `{level: ...}` form each level is its own nominal coverage.
- **Empty interval mapping** (CodeRabbit): already handled in d4899a3 — the dict branch now fully owns the dict case (incl. `{}`), so it can't fall through to the tuple unpack.

One I kept by design: the two tolerances are intentionally different roles. The metric-layer `tolerance` (0.05) is the per-level *calibration gate* — "is this level honest enough to count toward the verdict and the sharpness proxy?" The score-layer `_REG_CALIBRATION_TOLERANCE` is the scoring slope mapping ICE to 0–100, mirroring how ECE is mapped for classification. Happy to align them if you'd prefer a single knob.

Full suite green (424).

Khanz9664:
@Whatsonyourmind

Thanks for putting this together and for continuing the RFC-first workflow.

I appreciate the detailed PR description and the follow-up responses to the automated reviews. From a high level, it looks like you've stayed aligned with the design we converged on in RFC #155 while keeping the implementation backward compatible.

I haven't had a chance to do a proper review of the implementation yet, and I'd like to give it the same level of attention as the previous regression Trust Score work rather than rushing through it. I'll be going through the architecture, scoring logic, API integration, and the new test coverage in detail before leaving review feedback.

I'll start reviewing this as soon as possible and will follow up on the PR thread once I've completed a full pass.

Thanks again for the continued work on the regression framework.

Khanz9664 (commented):
@Whatsonyourmind

Thanks for putting this together. I've now gone through the implementation alongside RFC #155, the test suite, and the previous design discussion.

Overall, I think this is another very strong implementation. The architecture follows the RFC closely, the backward compatibility story is excellent, and I particularly like how the implementation preserves the three regression trust dimensions without introducing overlap between them.

A few things stood out positively:

* The separation between Accuracy, Interval Calibration (ICE), and Informativeness remains clean.
* The calibration-conditioned sharpness proxy is a nice implementation of the "resolution given calibration" idea we discussed during the RFC. Restricting the sharpness calculation to calibrated levels avoids rewarding artificially narrow but dishonest intervals.
* The fallback behaviour for existing single-level interval workflows is seamless, so this extends the framework without breaking existing users.
* The additional tests cover both the new metrics and the Trust Score integration very thoroughly.

I only have one design question that I'd like to discuss before approving.

### Zero calibrated levels and weight redistribution

One scenario I noticed is when a user provides multi-level prediction intervals, but none of those levels pass the calibration gate.

In that situation:

* `sharpness_skill` becomes unavailable,
* no uncertainty-correlation metric is present,
* and the Informativeness dimension is dropped through the existing weight redistribution.

That is consistent with the current "missing evidence → redistribute" philosophy, but this case feels slightly different to me because the evidence isn't actually missing—the model supplied interval estimates, and the framework determined that none of them were sufficiently calibrated to evaluate sharpness.

Because of that, I'm wondering whether this situation should instead produce an Informativeness score of `0.0` rather than removing the dimension entirely. Conceptually, it feels closer to "the supplied uncertainty was unusable" than "no uncertainty information was available."

I'm not convinced this is necessarily the right answer, but I think it's worth discussing since it slightly changes how models with consistently miscalibrated intervals are treated by the weighted score.

One smaller maintainability suggestion: I'd also recommend adding a brief inline comment explaining why the calibration gate tolerance used by the metric layer intentionally differs from the ICE-to-score mapping tolerance in `trust_score.py`. We discussed this during the RFC, but having that rationale directly in the code would help prevent future contributors from trying to unify the two values unnecessarily.

Other than those points, I think this is a very thoughtful implementation that stays faithful to RFC #155 while keeping the regression framework clean and extensible.

Whatsonyourmind:
Thanks for the careful review — and for the "resolution given calibration" framing, that's exactly the property I was trying to isolate.

I've pushed the inline-comment suggestion in `5cef03d`: `_REG_CALIBRATION_TOLERANCE` (0.20, the ICE→sub-score ramp) now carries a note on why it's intentionally distinct from the 0.05 calibration gate in `multilevel_interval_coverage` — gate = strict per-level pass/fail for sharpness eligibility; ramp = graceful 0–100 mapping — so nobody tries to unify the two later.

On the design question (multi-level intervals supplied, zero levels pass the gate → Informativeness currently dropped): I think you're right that this shouldn't be treated the same as "no intervals," and I'd lean toward your `0.0`.

The one thing I weighed is double-counting — ICE already penalizes the miscalibration (the Interval-Calibration sub-score is low when ICE is high), so a flat `Informativeness = 0.0` adds a second hit for the same underlying failure. But I don't think it actually *is* double-counting, because the two dimensions answer different questions and `0.0` is a truthful answer to the second one:

- **ICE**: "are the stated probabilities honest?" → no.
- **Informativeness**: "given honest probabilities, how much *usable resolution* did the uncertainty deliver?" → when nothing clears the gate, the answer is genuinely *zero usable resolution*, which is what `0.0` says. That's distinct from "no uncertainty was supplied at all" (dimension absent).

So I'd implement it as **Informativeness = 0.0 gated precisely on `interval_present and n_calibrated_levels == 0`** (i.e. `sharpness_skill is None` *because* the gate rejected every level, not because intervals were absent). The point-only / no-intervals path keeps redistributing exactly as today. This also closes the subtle leniency you're pointing at: today the 0.30 informativeness weight partly redistributes onto Accuracy, so an accurate-but-miscalibrated model can end up scoring *better* than it should — pinning it to `0.0` keeps that weight as a real penalty rather than a partial reward.

Two things I'd fold in to keep it clean and auditable:

1. An explicit status in the breakdown — e.g. `informativeness_status: "unusable_uncertainty"` vs `"absent"` — so a consumer of the score can tell "0.0 because the supplied intervals were all miscalibrated" from "dropped because none were supplied." Keeps the score explainable rather than a bare number.
2. Scope: I'd apply the `0.0` rule only to the multi-level sharpness path (where we actually had levels to evaluate and none qualified). The single-level PICP path falls back to the error-variance correlation and doesn't carry the same "we tried and it was unusable" signal, so I'd leave it on the redistribute path for now — happy to make them symmetric if you'd rather.

If that matches your intent I'll push it (small, self-contained, with tests for the zero-calibrated case and the status flag). Or if you'd prefer the informativeness weight redistributed to ICE-only instead of a hard `0.0`, that's a one-liner too — your call on which better fits the framework's philosophy.

Khanz9664 (commented):
@Whatsonyourmind

Thanks for the thoughtful response. I think we're aligned.

I agree with the distinction you've drawn between calibration and informativeness. ICE answers whether the reported uncertainty is trustworthy, while the informativeness dimension answers whether that trustworthy uncertainty actually provides useful resolution. In the "zero calibrated levels" case, those are two separate failures rather than the same one being counted twice, so I think treating Informativeness as `0.0` is a better reflection of what the model actually delivered than redistributing the weight.

I also like the idea of explicitly surfacing an `informativeness_status` (e.g. `unusable_uncertainty` vs `absent`). That keeps the score explainable and allows downstream consumers to distinguish between "no uncertainty was provided" and "uncertainty was provided but none of it was usable."

Limiting this behavior to the multi-level path also seems like the right scope for now, since that's the only case where we have enough evidence to conclude that the supplied uncertainty failed the calibration gate.

I think that would close the remaining design question for me. Once those changes are in, I'll go back through the updated implementation and expect to be in a position to approve.

Whatsonyourmind:
Pushed the zero-calibrated handling in `eee1949`. Summary of what's now in:

**1. `Informativeness = 0.0` for unusable multi-level uncertainty.** When multi-level intervals are supplied, every level fails the calibration gate (`sharpness_skill is None`, `n_calibrated_levels == 0`), **and** there's no error-variance-correlation fallback, the dimension is now scored a truthful `0.0` and kept (its 0.30 weight becomes a real penalty) rather than dropped-and-redistributed. This is exactly the "supplied but unusable" case you flagged.

**2. `informativeness_status` on `TrustScoreResult`** — `"present"` / `"unusable_uncertainty"` / `"absent"` (and `None` for classification). So a consumer can tell "0.0 because the intervals were all miscalibrated" from "dropped because none were supplied." It's also surfaced in `__str__` for the unusable case.

**3. Scoped to the multi-level path** (`n_levels >= 2`). The single-level PICP path has no "we tried and it was unusable" signal, so it keeps redistributing.

One precedence decision worth your eyes: I placed the `0.0` branch **after** the error-variance-correlation fallback, so if a usable correlation *is* present it still wins (matches your "no uncertainty-correlation metric is present" framing and keeps `test_falls_back_to_correlation_when_no_sharpness` intact). The `0.0` only fires when nothing else can score the dimension. If you'd rather have unusable multi-level intervals hard-pin to `0.0` even when a correlation exists, that's a one-line reorder — your call on which better fits the framework's philosophy.

Added 6 tests (`TestMultilevelIntervalCalibrationRFC155`): the `0.0` case + kept-weight, the unusable-scores-below-usable invariant, the correlation-fallback precedence guard, the single-level scoping guard, and the status-flag values (incl. `None` for classification). Full suite green (430 passed), `ruff check`/`ruff format --check` clean, and `mypy trustlens/trust_score.py --follow-imports=skip` reports no issues.

Ready for your final pass whenever you have a moment.

Khanz9664 (approved):
@Whatsonyourmind

Thanks for following up so quickly on the review comments.

I've gone through the latest changes, and I think this resolves the remaining design question nicely.

I particularly like the distinction between `"absent"` and `"unusable_uncertainty"` through `informativeness_status`. That keeps the Trust Score explainable and allows downstream consumers to distinguish between missing uncertainty evidence and supplied uncertainty that failed the calibration gate. I also agree with the precedence you've implemented—using a valid error-variance correlation when available before falling back to the `0.0` informativeness case feels like the right interpretation of the available evidence.

The new tests cover the intended semantics well, especially the zero-calibrated case, the retained-weight behavior, the precedence guard, and the multi-level scoping. Combined with the earlier documentation update around the separate calibration tolerances, I think this leaves the implementation in a much stronger and more maintainable state.

Thanks again for being receptive to the feedback throughout this RFC and implementation cycle. The discussion around these edge cases has made the regression framework considerably stronger, and I think this PR provides a solid extension to the Trust Score while remaining faithful to the architecture we established in RFC #155.

**Approved.**

PR #163, opened 2026-07-03
feat(calibration): add Maximum Calibration Error (MCE) via a shared ECE/MCE binning core

### LINKED ISSUE #134: [FEATURE]: Add Maximum Calibration Error (MCE) metric to the Calibration module
### Problem Statement

TrustLens currently computes **Expected Calibration Error (ECE)** — the *average* miscalibration across confidence bins. But ECE can hide dangerous outlier bins where the model is severely miscalibrated in a narrow confidence range.

**Maximum Calibration Error (MCE)** is the *worst-case* calibration gap across all bins. It is the number a safety engineer cares about: not how bad calibration is on average, but how bad it gets in the worst case.

The two metrics are complementary:
- ECE = 0.04 (looks fine) but MCE = 0.31 (one bin is catastrophically miscalibrated) → model should still fail a safety audit
- MCE is already listed in the roadmap and is a standard metric from the calibration literature (Naeini et al., 2015)

Currently there is no way to compute MCE through TrustLens, and it does not appear in `TrustReport` output.

---

### Proposed Solution

**Step 1 — Implement `maximum_calibration_error()` in `trustlens/metrics/calibration.py`**

The function signature should mirror `expected_calibration_error()` exactly so it's a drop-in companion:

```python
def maximum_calibration_error(
    y_true: np.ndarray,
    y_prob: np.ndarray,
    n_bins: int = 10,
) -> float:
    """
    Compute Maximum Calibration Error (MCE).

    MCE is the worst-case calibration gap across all confidence bins.
    It captures the most dangerous miscalibration region that ECE averages away.

    Parameters
    ----------
    y_true : np.ndarray
        Binary ground-truth labels (0 or 1), shape (n_samples,).
    y_prob : np.ndarray
        Predicted probabilities for the positive class, shape (n_samples,).
    n_bins : int
        Number of equal-width confidence bins. Default 10.

    Returns
    -------
    float
        MCE in [0, 1]. Lower is better. 0.0 means perfect calibration in every bin.

    Notes
    -----
    MCE >= ECE always holds. A model can have low ECE but high MCE when
    miscalibration is concentrated in a single bin.
    """
    y_true = np.asarray(y_true, dtype=float)
    y_prob = np.asarray(y_prob, dtype=float)

    bin_edges = np.linspace(0.0, 1.0, n_bins + 1)
    max_gap = 0.0

    for lo, hi in zip(bin_edges[:-1], bin_edges[1:]):
        mask = (y_prob >= lo) & (y_prob < hi)
        if mask.sum() == 0:
            continue
        accuracy = y_true[mask].mean()
        confidence = y_prob[mask].mean()
        max_gap = max(max_gap, abs(accuracy - confidence))

    return float(max_gap)
```

**Step 2 — Export from `trustlens/metrics/__init__.py`**

Add `maximum_calibration_error` to the `__all__` list alongside `expected_calibration_error`.

**Step 3 — Wire into `trustlens/core/pipeline.py`**

Inside the calibration block (around line 168), add:

```python
from trustlens.metrics.calibration import maximum_calibration_error
# ...
results["calibration"]["mce"] = maximum_calibration_error(y_true, y_prob_pos)
```

Note: MCE is only meaningful for binary calibration (same constraint as the current ECE path). The multiclass path (top-label ECE, line ~158) should also receive an MCE computed on `correct_mask` vs `confidences`.

**Step 4 — Surface in `TrustReport.show()`**

Add MCE to the calibration summary line so it appears alongside ECE in console output.

**Step 5 — Write tests in `tests/test_calibration.py`**

Required test cases:

1. `test_mce_perfect_calibration_is_zero` — y_prob perfectly predicts y_true → MCE = 0.0
2. `test_mce_geq_ece` — for any random input, MCE ≥ ECE (mathematical property)
3. `test_mce_worst_bin_dominates` — construct a case where one bin has a large gap and verify MCE captures it while ECE would understate it
4. `test_mce_empty_bins_handled` — skewed probability distributions that leave some bins empty do not raise errors
5. `test_mce_in_pipeline_results` — run `analyze()` end-to-end and confirm `results["calibration"]["mce"]` is present and is a float in [0, 1]
6. `test_mce_quantile_bin_collapse` — verifies MCE remains stable when quantile binning produces fewer effective bins than requested.
7. `test_mce_empty_input_raises` — empty arrays raise `ValueError`.
8. `test_mce_matches_ece_binning_logic` — validates MCE uses the same sample assignment strategy as ECE.

**Step 6 — Update docs**

- `docs/metrics/calibration.md` (or equivalent): add MCE alongside ECE
- `CHANGELOG.md` under `[Unreleased]`

---

### Alternatives Considered

MCE was previously out of scope while ECE was being stabilized. ECE is now stable and well-tested, making this the right time to add the companion metric.

---

### Additional Context

**Difficulty:** 🟡 Intermediate

**Why MCE matters:** ECE averages calibration error across all bins. A model with ECE=0.04 can still have MCE=0.35 if one confidence region is severely wrong. MCE is the safety-critical complement to ECE.

**Reference:** Naeini, M.P., Cooper, G., Hauskrecht, M. (2015). "Obtaining Well Calibrated Probabilities Using Bayesian Binning." AAAI.

| Reference file | Purpose |
|---|---|
| `trustlens/metrics/calibration.py` | Add the function here |
| `trustlens/core/pipeline.py` | Wire it into the pipeline |
| `tests/test_calibration.py` | Add tests here |
| `CONTRIBUTING.md` → Section 5 | "Adding a New Metric" guide with MCE as the worked example |

**PR checklist:**
- [ ] `maximum_calibration_error()` implemented in `metrics/calibration.py`
- [ ] Exported from `metrics/__init__.py`
- [ ] Wired into `pipeline.py` calibration block
- [ ] `report.show()` displays MCE alongside ECE
- [ ] ≥ 5 tests in `tests/test_calibration.py`
- [ ] `CHANGELOG.md` updated

---

### Implementation Notes

To ensure consistency with the existing calibration module:

- `maximum_calibration_error()` should follow the same binning strategy and sample-assignment logic used by `expected_calibration_error()`.
- Contributors should avoid introducing a separate binning implementation for MCE that could diverge from ECE results.
- Empty confidence bins should be skipped, consistent with existing ECE behavior.
- Empty input arrays should raise a `ValueError` rather than returning a calibration score.

---

### Calibration Consistency Requirement

During discussion of this feature, @cschanhniem identified potential inconsistencies between:

- `expected_calibration_error()`
- `reliability_curve()`

In particular:

- Quantile binning can collapse bins when duplicate probability values occur.
- `reliability_curve()` currently uses a different bin-assignment strategy than ECE.

Any MCE implementation should verify behavior under:

- Quantile-bin collapse
- Highly concentrated probability distributions
- Extreme confidence values (0.0 and 1.0)

If calibration binning logic is modified, consistency across ECE, MCE, and reliability visualizations must be preserved.

Whatsonyourmind:
Implements #134 — `maximum_calibration_error()`, the worst-case companion to ECE.

## What's in

**The metric** (`trustlens/metrics/calibration.py`): largest per-bin |accuracy − confidence| gap, per Naeini et al. (2015). Signature mirrors `expected_calibration_error()` exactly, including the `uniform`/`quantile` strategies, so it's a drop-in companion.

**Consistency by construction.** The issue's implementation notes (and @cschanhniem's binning analysis) ask that MCE follow ECE's binning and sample-assignment logic exactly, without a second implementation that could diverge. Rather than copying the loop, I extracted ECE's existing binning into a shared core (`_binned_calibration_gaps`) that returns the per-bin `(weight, gap)` table; ECE is now the weighted sum over that table and MCE the max. Same edges, same half-open/closed-last-bin assignment, same empty-bin skipping, same quantile-edge deduplication — divergence is structurally impossible. ECE's arithmetic is unchanged (same accumulation order), and the characterization parity suite passes against the stored baselines, confirming bit-identical ECE results.

**Pipeline wiring** (`trustlens/core/pipeline.py`): `results["calibration"]["mce"]` on both the binary path (`y_true` vs `y_prob_pos`) and the multiclass top-label path (`correct_mask` vs `confidences`), per the issue's note.

**Report output**: `TrustReport.show()` renders every scalar calibration key through the generic module printer, so MCE appears alongside ECE in the Calibration Analysis section ("- Mce: 0.xxxx") with no report-code change needed.

**Validation**: empty inputs raise `ValueError`; shape mismatches get the same clear error style as `brier_score`.

## Tests (12 new, all spec cases covered)

- `test_mce_perfect_calibration_is_zero`
- `test_mce_geq_ece` — both strategies (mathematical property)
- `test_mce_worst_bin_dominates` — exact construction: ECE = 0.095 "looks fine", MCE = 0.95 exposes the catastrophic bin
- `test_mce_empty_bins_handled` — skewed distributions
- `test_mce_in_pipeline_results` — end-to-end `analyze()`; also asserts MCE ≥ ECE on the live result
- `test_mce_quantile_bin_collapse` — constant and heavily-duplicated probabilities
- `test_mce_empty_input_raises`
- `test_mce_matches_ece_binning_logic` — with `n_bins=1` the weighted average and the max are the same number, so ECE == MCE exactly; any divergence in edge computation or sample assignment breaks this identity
- `test_mce_extreme_probabilities_included` — p = 1.0 lands in the closed last bin, exactly like ECE
- shape-mismatch / invalid-strategy / returns-float

Full suite: 442 passed. `ruff check` + `ruff format --check` clean. The one export-consistency test (`test_metrics_all_exports.py`) was extended to include the new symbol.

## Notes

- `docs/metrics/calibration.md` gains an MCE bullet (the API Reference section picks the function up via automodule); CHANGELOG updated under Unreleased.
- Scope kept to the metric + wiring: no insight/pattern rules reference MCE yet (e.g. a "low ECE, high MCE" observation in `_generate_insights()` could be a natural follow-up if you want one).

## Summary by CodeRabbit

* **New Features**
  * Added Maximum Calibration Error (MCE) alongside existing calibration metrics.
  * MCE is now computed and included in calibration analysis results and report summaries for both binary and multiclass workflows.
* **Documentation**
  * Updated calibration documentation with an interpretation of ECE vs worst-case bin miscalibration (MCE).
  * Updated the changelog to reflect the new metric’s addition and public availability.


Khanz9664:
@Whatsonyourmind 

Thanks for taking this on and for putting together what looks like a very thorough implementation.

I appreciate the effort that went into this PR, especially the architectural decision to build MCE on top of a shared internal binning core rather than introducing a second implementation. That aligns well with one of the primary goals discussed on Issue #134—keeping ECE and MCE consistent by construction instead of relying on two independent code paths. 

I also noticed you've incorporated the edge cases that came up during the issue discussion (quantile-bin collapse, extreme probabilities, pipeline integration, etc.), which I appreciate.

I've started reviewing the implementation in detail now. As with the recent regression RFC work, I'll be looking not only at the correctness of the implementation itself but also at how well it preserves the existing calibration architecture, maintains backward compatibility, and behaves across edge cases.

If I have any questions or suggestions during the review, I'll leave them here so we can discuss them before moving toward a merge. Thanks again for the comprehensive work on this one.

Khanz9664 (changes requested):
@Whatsonyourmind 

Thanks for taking this on. I've finished a detailed review of the implementation.

Overall, I think this is a very well-structured PR. The decision to extract the shared `_binned_calibration_gaps()` helper instead of implementing MCE independently is exactly the direction I was hoping to see. It guarantees that ECE and MCE will always share identical binning semantics, which was one of the primary goals discussed in Issue #134, and it removes the risk of the two metrics drifting apart over time.

I also appreciate the thorough testing you've added. The mathematical property tests (`MCE >= ECE`), the catastrophic-bin example, the pipeline integration, and the export/documentation updates all make this feel like a complete feature rather than just a metric implementation.

I do have a few items I'd like to see addressed before merging.

### 1. Fully collapsed quantile bins (major)

There is one edge case in `_binned_calibration_gaps()` that I think needs handling.

When using `strategy="quantile"`, it's possible for

```python
bin_edges = np.unique(bin_edges)
```

to collapse down to a single value if every predicted probability is identical.

In that situation,

```python
zip(bin_edges[:-1], bin_edges[1:])
```

produces no bins, so `_binned_calibration_gaps()` returns empty `weights` and `gaps`.

That means

* `expected_calibration_error()` returns `0.0`
* `maximum_calibration_error()` also returns `0.0`

even if the constant predictor is badly miscalibrated.

The current `test_mce_quantile_bin_collapse()` doesn't expose this because it uses a perfectly calibrated constant predictor (`p = 0.5`, empirical accuracy = `0.5`).

I'd recommend explicitly handling the `len(bin_edges) == 1` case by treating the entire dataset as a single active bin and adding a regression test with something like

```python
y_true = np.ones(...)
y_prob = np.full(..., 0.5)
```

where the expected calibration gap is non-zero.

---

### 2. Input validation in the shared helper (medium)

Since `_binned_calibration_gaps()` is now the common implementation used by both ECE and MCE, I think it's worth validating its inputs directly rather than relying on callers.

Two cases stood out:

* reject non-1D arrays
* reject `n_bins < 1`

For example, same-shaped 2D arrays currently pass through and `len(y_true)` becomes the denominator, which can produce incorrect weights.

Similarly, `n_bins = 0` currently falls into the collapsed-bin behavior instead of producing a clear error.

Adding explicit validation here makes the shared core more robust.

---

### 3. Multiclass pipeline integration test (minor)

The pipeline wiring correctly adds MCE to both the binary and multiclass paths, but the integration test currently exercises only the binary branch.

I'd like to see a small multiclass end-to-end test (`n_classes >= 3`) verifying that the top-label calibration path exposes

```python
results["calibration"]["mce"]
```

correctly.

The metric implementation itself is shared, so I'm not expecting problems here—this is mainly to cover the pipeline wiring.

---

### 4. CONTRIBUTING example (minor)

One small documentation observation:

`CONTRIBUTING.md` still uses "adding Maximum Calibration Error (MCE)" as the hypothetical example for extending TrustLens.

Now that MCE actually exists, that example could become confusing for future contributors.

I'd suggest replacing it with another fictional metric so the guide remains forward-looking.

---

Other than those points, I think this is an excellent implementation. The shared binning core is a nice architectural improvement, the API remains fully backward compatible, and the overall quality of the implementation is high. Once these edge cases are addressed, I think this will be in great shape for merging.

Whatsonyourmind:
Thanks for the detailed review — all four addressed in `ba4d912`.

**1. Fully collapsed quantile bins (major).** You're exactly right, and the existing test masked it by using a perfectly calibrated constant (`p=0.5`, acc `0.5`, gap `0`). `_binned_calibration_gaps()` now detects the collapsed-edge case (`len(bin_edges) < 2`) and falls back to treating the whole dataset as a single active bin, so a miscalibrated constant predictor reports its true gap instead of `0.0`. New regression test uses your suggested shape — `y_true = ones(100)`, `y_prob = full(100, 0.5)` — and asserts **both** ECE and MCE return `0.5` (not `0.0`).

**2. Input validation in the shared core (medium).** Added directly to `_binned_calibration_gaps()` since it's now the common path:
- non-1D arrays are rejected (the same-shaped-2D case you flagged, which corrupted the `len(y_true)` denominator for *both* ECE and MCE — ECE had no shape guard at all before this);
- `n_bins < 1` now raises instead of silently falling into the degenerate-bin path.

Putting the checks in the shared core means ECE and MCE stay consistent on malformed input too, not just on the happy path. Covered by `test_binned_core_rejects_non_1d_and_bad_n_bins`.

**3. Multiclass pipeline integration test (minor).** Added `test_mce_in_pipeline_results_multiclass` — an `n_classes=3` end-to-end `analyze()` run asserting `results["calibration"]["mce"]` on the top-label branch. As you expected, the shared metric needed no change; this just locks the multiclass wiring.

**4. CONTRIBUTING example (minor).** Good catch — replaced the now-real MCE walkthrough with a hypothetical *Signed Calibration Bias (SCB)* metric (population-weighted **signed** confidence−accuracy). I deliberately picked a signed/directional metric so it reads as clearly distinct from the real unsigned ECE/MCE and won't create the same "wait, does this exist?" confusion later.

Full suite is green locally (445 passed), mypy and ruff clean. Ready for another look when you have a moment.

Khanz9664 (approved):
@Whatsonyourmind 

Thanks for addressing the review so thoroughly. I've gone through the latest changes, and I think this is in a good place now.

The collapsed quantile-bin handling is exactly what I was hoping to see. Treating the fully collapsed case as a single active bin preserves the correct calibration signal instead of silently reporting `0.0`, and the new regression test using a miscalibrated constant predictor nicely captures the original failure mode.

Moving the validation into the shared `_binned_calibration_gaps()` helper is also the right architectural choice. Since both ECE and MCE now depend on the same implementation, validating the common path guarantees the two metrics remain consistent on malformed inputs as well as on valid ones.

I also appreciate the additional multiclass pipeline integration test. The metric implementation itself was already shared, but having an end-to-end test for the multiclass branch closes an important gap in coverage.

Finally, thanks for updating the `CONTRIBUTING.md` example. Replacing the MCE walkthrough with a fictional metric avoids future confusion and keeps the documentation forward-looking.

Overall, this is a well-executed implementation. The shared binning core simplifies maintenance, the regression tests strengthen confidence in the edge cases, and the overall design remains clean and backward compatible.

Nice work—I'm happy with these changes. Thanks again for the thoughtful implementation and the quick turnaround on the review feedback.

Whatsonyourmind:
Thanks for the merge, and for the careful review that got it there — the collapsed-quantile-bin catch in particular was a real latent bug (both ECE and MCE silently reporting 0.0 for a miscalibrated constant predictor), so I'm glad it surfaced before release rather than after. Happy to have MCE sharing the binning core with ECE now.

PR #164, opened 2026-07-05
feat(metrics): conformal prediction diagnostics (metrics module) — implements RFC #157 (PR 1/2)

### LINKED ISSUE #157: [FEATURE]: Conformal Prediction Diagnostics (Set Size & Conditional Coverage)
**Motivation**
While TrustLens offers excellent standard calibration metrics (ECE, Brier Score, and now PICP for regression), the industry standard for robust uncertainty quantification is shifting toward **Conformal Prediction** (CP). 

Conformal classifiers output *prediction sets* rather than single point predictions (e.g., `{Cat, Dog}` instead of just `Cat`). TrustLens currently lacks a native way to evaluate the quality of these prediction sets, leaving a gap for modern safety-critical deployments.

**Proposed Solution**
Add a new `conformal` dimension (or augment the existing uncertainty/calibration modules) that accepts prediction sets and evaluates:
1. **Marginal Coverage**: Does the true label fall into the prediction set $1-\alpha$ percent of the time?
2. **Conditional Coverage**: Is the coverage maintained across different classes and confidence bins?
3. **Average Set Size**: Are the sets actually informative? (A set containing all possible classes has 100% coverage but 0 informativeness).

**Proposed API**
```python
# y_pred_sets is a list of lists or binary matrix indicating set membership
report = TrustLens.evaluate(
    y_true=y_true,
    y_pred_sets=prediction_sets, # New parameter for conformal sets
    nominal_coverage=0.90
)

# Output would surface in the calibration/informativeness report:
# - Marginal Coverage: 90.5% (Target: 90%)
# - Average Set Size: 1.4 classes
# - Singleton Rate: 85% (Percentage of time the set size is exactly 1)
```

**Open Questions**
- Should this live as a new top-level dimension, or be folded into the existing `Calibration` and `Failure` modules?
- Should we accept sets as lists of labels, or as a binary dense matrix of shape `(n_samples, n_classes)`? Matrix form is likely much faster for NumPy broadcasting.

Whatsonyourmind:
Implements the metrics layer of the conformal-prediction-diagnostics RFC in #157, along the split we agreed there.

## Scope of this PR (1/2)

Just the numpy-only metrics module + tests + docs — no pipeline/report/`trust_score.py` changes. This deliberately isolates the **function interfaces and semantics** so they can be locked before any wiring, per your "settle the design first" preference.

Why split it here rather than bundle like MCE (#163): MCE reused the existing `y_prob` input, so it could wire into the pipeline in one PR. Conformal needs a **new `y_pred_sets` public input** on `analyze()` — a bigger public-surface decision. Keeping that (plus the report panel) as PR 2 gives the new API its own focused review. **Phase 2 Trust Score integration stays deferred** per your note — nothing in this PR touches scoring.

## What's in — `trustlens/metrics/conformal.py`

Same conventions as `metrics/calibration.py` (pure functions, plain floats/dicts, exported in `metrics/__init__.py`). Canonical internal form is an `(n, K)` boolean membership matrix.

- **`to_membership_matrix(y_pred_sets, n_classes=None)`** — normalises either an `(n, K)` 0/1 matrix or a ragged list of per-sample label lists to a boolean set matrix.
- **`marginal_coverage`** — overall `P(y ∈ C(x))`.
- **`class_conditional_coverage`** — per-class coverage + worst-class gap (the class where the guarantee quietly fails).
- **`size_stratified_coverage`** — coverage within each set-size stratum (Angelopoulos & Bates SSC), with a `min_stratum` (default 20) small-stratum guard so a tiny stratum can't dominate the violation.
- **`set_size_summary`** — avg size, singleton rate, `size_efficiency` (`1` = all singletons, `0` = all-`K`), empty rate.
- **`conformal_diagnostics`** — single orchestrated report. When `nominal_coverage` is absent, every gap/violation field is `None` (never a silent `0`) so a missing target is unmistakable.

**Method-agnostic:** TrustLens evaluates prediction sets from any source (LAC/APS/RAPS/Mondrian/…); it does not generate them and takes no CP-library dependency — the same stance the regression path takes toward interval methods.

**Scope discipline:** these are *approximate* conditional diagnostics — exact distribution-free conditional coverage is impossible in finite samples (Vovk 2012; Barber et al. 2021), so the docs frame class/size-conditional numbers as failure-surfacing signals read alongside their per-group counts, not guarantees.

## Tests — `tests/test_conformal.py`

37 tests on hand-built sets with known answers: marginal / worst-class / SSC / size on constructed cases, the all-`K` and all-singleton extremes, empty-set handling (counted as a miss), the `min_stratum` guard, the nominal-unknown path (gaps → `None`), and the headline **marginal-vs-SSC divergence** case (high marginal while a stratum under-covers). `test_metrics_all_exports.py` extended for the new symbols.

Full run: **41 passed** (37 + 4 export). `ruff check` + `ruff format --check` clean. Docs: `docs/metrics/conformal.md` (automodule picks up the API); `CHANGELOG.md` under Unreleased.

## Next (PR 2)

`analyze()` activation via `y_pred_sets` (block emitted iff sets supplied; coexists with `y_prob`) + the Conformal Prediction panel under Calibration Analysis. Then, separately and only once we've seen the diagnostics on real models, the Phase 2 scoring discussion.

Refs #157.

## Summary by CodeRabbit

* **New Features**
  * Added conformal prediction diagnostics for prediction sets, including marginal, class-conditional (worst-class), and size-stratified coverage plus set-size summary metrics.
  * Added a single consolidated diagnostics report with optional nominal coverage gap/violation fields.
* **Documentation**
  * Added a full “Conformal Prediction Diagnostics” documentation page and updated the changelog.
* **Tests**
  * Added end-to-end and helper-function tests covering validation, edge cases, and expected metric outputs.
* **Chores**
  * Exposed the new conformal metrics at the metrics package level for easier access.


Whatsonyourmind:
Addressing the automated review (llamapreview P1 + CodeRabbit) on out-of-range `y_true` labels in `_coverage_vector` / `_check_y_true`.

I looked at raising `ValueError` there, but it's actually the wrong call — and it breaks a real, intended behavior. When a class is rare/hard and the predictor **never places it in any prediction set**, the inferred `K` (class span of the sets) won't include it, so a legitimate true label ends up `>= K`. That sample is a **genuine miss** — the class truly is never covered — and `class_conditional_coverage` reporting **0.0 coverage for that class is exactly the headline failure signal these diagnostics exist to surface**, not an input error. Raising would turn the most important finding into an exception. (Locally, an always-raise implementation breaks `test_empty_set_is_a_miss` and friends, which encode this.)

So I've taken the reviewers' "document explicitly if intentional" path instead (commit 1974b72):
- expanded the `_coverage_vector` docstring to state that a label `>= K` is an intentional miss, with the rationale (negatives are still rejected by `_check_y_true`);
- added a note to `class_conditional_coverage` that a class present in `y_true` but in no set surfaces as 0.0;
- added `test_class_never_in_any_set_is_zero_coverage` to lock it (now 38 tests, mypy + ruff clean).

Two things worth your call for later PRs, both tied to the same root (`K` is inferred from the sets, not `y_true`):
1. For the entrypoints that take an explicit `n_classes` (`class_conditional_coverage`, `conformal_diagnostics`), a `y_true` label `>= n_classes` **is** an unambiguous contract violation and could reasonably raise there — happy to add that if you'd like, since it can't collide with the never-covered-class case.
2. This is the same matrix-vs-label-list disambiguation I flagged in the RFC: `[[0], [1]]` reads as a `(2, 1)` matrix rather than two singleton label-lists. If you'd prefer a different default there, it changes the `K` inference and I'll fold it in.

Khanz9664 (changes requested):
@Whatsonyourmind 

Thanks for putting this together. I've now finished a full review of the PR.

Overall, I think this is a very strong implementation. Splitting the work into a metrics-only PR before introducing the public `analyze()` API was the right design decision, and it keeps the review focused on the mathematical semantics first. The module is cleanly structured, the documentation is excellent, and I especially appreciate the extensive hand-constructed test cases—they give a lot of confidence that the implementations match the intended conformal metrics rather than only satisfying property-based tests.

I also agree with your reasoning in the discussion around the automated review. Treating a true label outside the inferred class span as a coverage miss (rather than immediately raising) is the correct behavior when `K` is inferred from the prediction sets. If a predictor never includes a particular class in any prediction set, surfacing that as 0% coverage is exactly the diagnostic this module is intended to expose, and your additional documentation and regression test make that intent much clearer.

There are still a few things I'd like to see addressed before I'm comfortable merging.

---

### 1. Ambiguous rectangular list-of-lists parsing

One case still worries me:

```python
[[0], [1]]
```

or more generally any rectangular list-of-lists containing only `0`s and `1`s.

This representation is inherently ambiguous.

It could legitimately mean:

* a membership matrix

or

* two singleton prediction sets represented as label lists.

Today the parser silently chooses the matrix interpretation.

The problem isn't that either interpretation is wrong—it's that both are valid, and the user receives completely different coverage metrics without any indication that their input was interpreted differently than intended.

NumPy arrays are already unambiguous, so I think the ambiguity only exists for native Python lists.

I'd therefore prefer the API to reject this specific ambiguous case rather than silently picking one interpretation.

Something along the lines of:

```text
Ambiguous input: a rectangular list containing only 0/1 values could represent
either a membership matrix or label lists. Please provide n_classes or pass
a NumPy array to disambiguate.
```

would make the API much safer.

---

### 2. Explicit `n_classes` should enforce the declared contract

I also wanted to follow up on something you mentioned in your own comment.

I completely agree that

* inferred `K` → out-of-span labels should remain coverage misses

but

* explicit `n_classes`

is a different situation.

If a caller explicitly declares

```python
n_classes = 10
```

while `y_true` contains class `15`, that is no longer a diagnostic signal—it's a contradiction of the user's own declared contract.

I think that should raise a `ValueError` rather than silently treating it as a miss.

That preserves the intended "never-covered class" behavior when `K` is inferred, while still protecting users from accidental data inconsistencies when they've explicitly specified the class space.

---

### 3. API consistency across the module

One small API consistency point.

Currently

* `class_conditional_coverage`
* `set_size_summary`
* `conformal_diagnostics`

accept `n_classes`

whereas

* `marginal_coverage`
* `size_stratified_coverage`

do not.

Even though `n_classes` doesn't change the mathematical definitions of marginal coverage or SSC, passing it through to `to_membership_matrix()` would provide the same validation and disambiguation behavior across every public entry point.

I think having every public conformal metric expose the same optional `n_classes` parameter would make the module feel more consistent and reduce surprises for users.

---

Other than those points, I think this is an excellent foundation for the conformal diagnostics work. The separation of concerns is clean, the documentation explains both the statistical motivation and the implementation choices well, and the test suite gives me a lot of confidence in the numerical correctness of the metrics.

Once these remaining API edge cases are addressed, I think this will be in great shape to merge.

Whatsonyourmind:
Thanks for the thorough review — all three addressed in `3bbd2f0` (rebased onto `main` now that #163 has merged, so this sits cleanly on the shared binning core).

**1. Ambiguous rectangular 0/1 list-of-lists — now rejected.** You're right that this is the one genuinely ambiguous case, and only for native lists (NumPy arrays already disambiguate: 2D → matrix, 1D object → label lists). `to_membership_matrix` now raises for a rectangular native list whose entries are all 0/1 unless the caller resolves it:

```
Ambiguous input: a rectangular list containing only 0/1 values could represent
either an (n, K) membership matrix or a list of label lists. Pass n_classes to
disambiguate (a matrix has exactly n_classes columns), or pass a NumPy array
(2D = matrix, 1D object array = label lists).
```

When `n_classes` *is* given, it disambiguates by width: `width == n_classes` → matrix, any other width → label lists (so `[[0],[1]]` with `n_classes=3` is unambiguously the two singleton sets `{0}`, `{1}`). Non-0/1 or ragged native lists remain unambiguous label lists as before.

**2. Explicit `n_classes` is a contract — now raises on violation.** Exactly the distinction you drew. When `K` is *inferred*, an out-of-span `y_true` label stays a coverage miss (the "class never predicted → 0% coverage" diagnostic, unchanged). When the caller *declares* `n_classes`, a `y_true` label outside `[0, n_classes)` is a contradiction of their own contract and now raises:

```
y_true contains label 15, outside the declared class space n_classes=10
(valid labels are 0..9). With an explicit n_classes this is a data-consistency
error; omit n_classes to treat unseen classes as coverage misses instead.
```

Implemented as a small `_enforce_declared_class_space` helper called right after `y_true` validation, so the two regimes stay cleanly separated.

**3. API consistency — `n_classes` everywhere.** `marginal_coverage` and `size_stratified_coverage` now accept the same optional `n_classes` as the other three entry points and route it through `to_membership_matrix`, so every public metric gets identical validation and disambiguation (and the contract check from point 2). Signatures stay backward compatible (`n_classes` is the last, optional arg).

Tests: the conformal suite is now 49 (added the ambiguity raise, all four disambiguation paths, the explicit-contract enforcement across all entry points, and the `n_classes` passthrough). Docs updated with the disambiguation rule and the contract semantics. Full suite **494 passed**, mypy and ruff clean. Ready for another look.

Khanz9664 (approved):
@Whatsonyourmind 

Thanks for the quick turnaround on the review feedback. I went through the updated implementation and I'm happy with the changes.

All three requested items have been addressed:

* The ambiguous rectangular native `0/1` list case is now explicitly rejected unless the caller disambiguates via `n_classes` or a NumPy array. This removes what was previously the biggest source of silent misinterpretation.
* Treating an explicit `n_classes` as a contract is a nice improvement. Separating the inferred-`K` diagnostic path from the explicit validation path via `_enforce_declared_class_space()` keeps the behavior both predictable and easy to maintain.
* Adding `n_classes` to the remaining public entry points makes the conformal API consistent across the entire module.

I also appreciate the additional documentation explaining the ambiguity rules and the explicit-vs-inferred class-space semantics, as well as the expanded test coverage around these behaviors.

The metrics module now feels well-scoped, internally consistent, and provides a solid foundation for the upcoming pipeline/report integration in the next phase of the conformal prediction work.

Nice work—I'm happy to approve this.

Khanz9664:
@Whatsonyourmind 

Thanks again for the work on PR #164. I've merged it after reviewing the implementation and running a round of maintainer-level validation on the metrics module. Everything looked solid—the ambiguity handling, explicit `n_classes` contract, API consistency, and the additional test coverage all landed well.

With the metrics layer now in place, I think we're in a good position to move on to **PR 2/2**.

My preference is to keep this PR focused on integration rather than introducing new metric semantics. The scope I'd suggest is:

* Wire the conformal diagnostics into the evaluation pipeline behind the new `y_pred_sets` input.
* Surface the diagnostics in the report structure and `TrustReport.show()` using the existing reporting conventions.
* Preserve the diagnostic-only behavior we agreed on in the RFC—no Trust Score changes yet.
* Add end-to-end integration tests covering the new pipeline path, report generation, and graceful degradation when `y_pred_sets` is absent.
* Update the user-facing documentation/examples for the new API.

Keeping this PR focused on plumbing and reporting will make it much easier to review independently from the later Trust Score integration.

Looking forward to reviewing it once it's ready.

Whatsonyourmind:
Thanks @Khanz9664 — glad the metrics module landed cleanly, and thanks for the careful pass on the ambiguity/contract edges.

Picking up **PR 2/2** now with exactly that scope:
- `y_pred_sets` as a new optional input on `analyze()`, validated through the same `to_membership_matrix` contract;
- the `results["calibration"]["conformal"]` sub-block emitted **iff** sets are supplied (coexisting cleanly with `y_prob` — ECE/Brier from probabilities, CP diagnostics from sets);
- a **Conformal Prediction** panel under Calibration Analysis, rendered through the existing generic scalar printer (the same trick MCE used in #163, so no bespoke report code);
- end-to-end integration tests: the pipeline path, the report render, and graceful degradation when `y_pred_sets` is absent (block simply not emitted);
- API docs + a short usage example.

Diagnostic-only — no `trust_score.py` changes; Phase 2 scoring stays parked exactly as we agreed.

One integration detail I'll settle in this PR since it surfaces in the report layer: the two `*_gap` fields carry opposite signs by design (`coverage_gap = marginal − nominal`, positive = *over*-coverage; `worst_class_gap = nominal − worst`, positive = *under*-coverage). Rather than change the merged metric semantics, I'll have the panel label them explicitly (over/under) so a reader can't misread the sign, and leave the raw fields untouched for downstream consumers. Say the word if you'd rather I rename in the report instead.

I'll open it against `main` once the integration tests are green.

Khanz9664:
@Whatsonyourmind 

Thanks for the detailed update. The proposed scope matches what I had in mind, and I like that you're keeping this PR focused purely on pipeline integration and reporting while leaving the Trust Score work for a later phase.

Regarding the integration detail, I agree with keeping the underlying metric semantics unchanged. Since the metrics module is already merged, I'd prefer not to rename the report fields or alter their meaning. Explicit labeling in the report to indicate whether a positive value represents over-coverage or under-coverage sounds like the right approach—it improves readability without introducing any API churn for downstream consumers.

Other than that, the planned scope looks good:

* `y_pred_sets` integrated into the evaluation pipeline.
* Conditional emission of the conformal diagnostics block alongside the existing probability-based calibration metrics.
* Report rendering following the existing scalar-report pattern.
* End-to-end integration tests and graceful degradation when conformal inputs are absent.
* Documentation and usage examples.

Looking forward to reviewing the PR once it's open.

PR #165, opened 2026-07-08
feat(conformal): pipeline + report integration via y_pred_sets (RFC #157, PR 2/2)

### LINKED ISSUE #157: [FEATURE]: Conformal Prediction Diagnostics (Set Size & Conditional Coverage)

Whatsonyourmind:
Implements **PR 2/2** of the conformal-prediction-diagnostics RFC (#157), following the merge of the metrics module in #164 and the scope you laid out. Pure plumbing + reporting — no new metric semantics, no Trust Score changes.

## Scope (exactly the five points from #164)

- **Wire the diagnostics into the pipeline behind a new `y_pred_sets` input.** `analyze(..., y_pred_sets=...)` accepts either an `(n, K)` membership matrix or ragged label lists (normalised by the merged `to_membership_matrix`). It's threaded through `_run_analysis_pipeline`, which computes `conformal_diagnostics()` into `results["calibration"]["conformal"]` iff sets are supplied. `confidence_level` doubles as the conformal nominal target (`1 − α`), so no extra public knob.
- **Surface it in the report.** A dedicated **Conformal Prediction** panel renders under Calibration Analysis in `TrustReport.show()`, in signal order: marginal coverage vs target → the two conditional diagnostics (worst-class gap, SSC violation) → informativeness (avg size, singleton rate, efficiency, empty rate). It uses the existing reporting conventions (the generic printer already hides the nested dict in non-verbose mode, so there's no messy double-render).
- **Diagnostic-only — no scoring.** `_calibration_score` reads only `brier_score`/`ece`, so the nested `conformal` block is inert; a dedicated test asserts the Trust Score and every sub-score are **identical** with and without sets. Phase-2 scoring stays parked exactly as we agreed.
- **End-to-end integration tests + graceful degradation.** `tests/test_conformal_integration.py` (11 tests): activation, coexistence with `y_prob`, nominal-from-`confidence_level`, list-of-lists input, the report panel present/absent, graceful skip on malformed sets (length mismatch → visible `status: "skipped"` block, analysis not aborted), conformal-without-probabilities, and score-invariance.
- **Docs/examples.** `docs/metrics/conformal.md` gains an "Using it through `analyze()`" section; `CHANGELOG.md` updated under Unreleased.

## One integration detail (flagged in the #164 thread)

The two `*_gap` fields carry opposite sign conventions by design — `coverage_gap = marginal − nominal` (positive = *over*-coverage) vs `worst_class_gap = nominal − worst` (positive = *under*-coverage). Rather than change the merged metric semantics, the panel labels each explicitly:

```
Conformal Prediction (prediction sets)
  Nominal coverage     : 0.9000
  Marginal coverage    : 0.9250  (over by 0.0250)
  Worst-class coverage : 0.8947  (under by 0.0053)
  SSC violation        : 0.0364
  Avg set size         : 1.1833
  Singleton rate       : 0.8167
  Size efficiency      : 0.9083
  Empty rate           : 0.0000
  (n=120, classes=3)
```

The raw fields keep their documented conventions for downstream consumers. Happy to rename in the report instead if you'd prefer.

## Behavior notes

- **Coexists with `y_prob`.** ECE/Brier/MCE from probabilities, coverage from sets; the block is even emitted when probabilities are withheld (it doesn't need them). Absent `y_pred_sets`, nothing changes.
- **Classification only** — `y_pred_sets` is ignored on the regression path.

## Tests

Full suite green: **476 passed** (excludes the optional xgboost/lightgbm/catboost backend tests, which need the `full` extra). `ruff check` + `ruff format --check` clean (0.15.x), mypy clean on the changed files.

## Checklist

- [x] Branch name is not `main` or `dev`
- [x] Tests pass locally (`pytest tests/test_conformal_integration.py`)
- [x] Code is formatted (`ruff format && ruff check`)
- [x] Code coverage is not reduced (new integration tests added)
- [x] New functions/classes have type annotations and docstrings
- [x] Docs updated (`docs/metrics/conformal.md`, `CHANGELOG.md`)
- [x] Licensed code from external sources is explicitly attributed (n/a — original implementation)

refs #157

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end conformal prediction diagnostics for classification via `analyze()` using `y_pred_sets`.
  * Reports now include a dedicated **Conformal Prediction** panel (and saved text output mirrors it).
  * `confidence_level` is used as the nominal target for these diagnostics (diagnostic-only; Trust Score unchanged).
* **Bug Fixes**
  * Malformed or empty prediction sets no longer halt analysis; they show a visible skipped diagnostics block.
  * Conformal diagnostics still appear even when probability-based calibration outputs are unavailable.
* **Documentation / Tests**
  * Added documentation and comprehensive integration tests covering activation, formatting, degradation, and report behavior.


Whatsonyourmind:
Addressed the automated review in `dda2dfd`:

**1. Text report omitted the panel (CodeRabbit) — fixed.** Good catch: `show()` rendered the conformal panel but `_generate_text_report()` (what `save(".txt")` uses) skipped it, since the generic text helper hides nested dicts. I extracted the panel into a single `_conformal_panel_lines()` builder that both `show()` (prints) and `_generate_text_report()` (buffers) call, so interactive and saved output are now identical. Added a test asserting the panel is present in the text report with sets and absent without.

**2. Bare `except Exception` (llamapreview P2) — narrowed.** Now `except (ValueError, TypeError)` — exactly the input-validation errors `conformal_diagnostics` / `to_membership_matrix` raise (length mismatch, all-empty, ambiguous 0/1 list, out-of-contract labels). A genuine bug in the metric now propagates instead of being silently downgraded to a skipped block.

**3. Hardcoded `5e-5` (llamapreview P2) — named.** Promoted to `_GAP_ON_TARGET_EPS` with a comment (it's the magnitude below which a gap rounds to `0.0000` at the panel's 4-decimal precision, so it reads as "on target").

**4. Positional-arg break (llamapreview P2) — not applicable.** `y_pred_sets` is inserted in the **keyword-only** block (after the `*` in `analyze()`), alongside `confidence_level`/`prediction_intervals`/etc., so there are no positional callers to break — all of these are keyword-only by design.

**5. `Optional[Any]` looseness (llamapreview P2) — intentional, left as-is.** `y_pred_sets` genuinely accepts three shapes — an `(n, K)` NumPy matrix, a NumPy object array of label lists, or a native ragged `list[list[int]]` — and the real validation/disambiguation lives in `to_membership_matrix` (which raises with actionable messages). A precise static union would either be misleadingly narrow or as permissive as `Any`; I'd rather keep the runtime contract as the single source of truth. Happy to tighten if you'd prefer a `Union` annotation.

Full suite **477 passed**, ruff + mypy clean. Nothing here touches scoring.

Khanz9664 (approved):
@Whatsonyourmind 

Thanks for the excellent work on this PR!

I've completed a thorough review of the implementation and also pulled the changes locally to validate the integration. The separation between the metrics layer (PR #164) and this pipeline/report integration remains clean, and the implementation stays well within the agreed RFC scope.

* The conformal diagnostics integrate cleanly into `analyze()` via `y_pred_sets` without affecting the existing probability-based calibration workflow.
* The diagnostics coexist correctly with ECE/Brier/MCE while remaining completely diagnostic-only, keeping the Trust Score unchanged as planned.
* Extracting the conformal panel into a shared helper for both `show()` and the generated text report is a nice improvement that keeps both output paths consistent.
* Graceful degradation for malformed prediction sets is well handled, and narrowing the exception handling to expected validation errors avoids masking genuine implementation bugs.
* The integration tests provide solid coverage for activation, coexistence, report rendering, graceful degradation, and score invariance.

I also verified the follow-up changes responding to the previous review:

* Shared report generation between interactive and saved text output.
* Narrowed exception handling (`ValueError`/`TypeError`) instead of a broad `Exception`.
* Named constant replacing the magic tolerance value for coverage-gap formatting.
* Documentation updated with end-to-end usage examples.

Overall, this completes the second phase of RFC #157 exactly as discussed. The architecture remains modular, the public API is intuitive, and the implementation is well tested without introducing unnecessary coupling or premature Trust Score integration.

Whatsonyourmind:
Thanks @Khanz9664 — appreciate the thorough review and the local validation pass. Glad the metrics/pipeline split (#164 → #165) held up cleanly and stayed strictly diagnostic-only; keeping the Trust Score untouched until we deliberately decide otherwise was the part I most wanted to get right.

That wraps #157 end-to-end. The one piece still deferred from RFC #155 is CRPS-proper for Uncertainty Informativeness — #161 shipped the calibration-conditioned sharpness proxy we settled on for v1, so full CRPS (with CRPSS against the climatology reference) is the natural next increment whenever you want to pick it back up. Happy to spec it RFC-first as before; no rush.

Khanz9664:
@Whatsonyourmind 

Thanks for all the work you've put into both phases of this feature. Splitting the work into a metrics-first PR followed by a dedicated integration PR made the review process much easier, and the end result is a clean, well-structured implementation. I also pulled everything locally after the merge and ran an additional round of maintainer-level validation beyond the CI suite, and it's great to see the design hold up well end-to-end.

I agree that RFC #157 is now in a good place. Keeping the conformal diagnostics purely informational in this first iteration was the right decision, and we now have a solid foundation if we decide to incorporate them into the Trust Score in the future.

Regarding RFC #155, I also agree that CRPS/CRPSS is the next logical enhancement for the regression uncertainty story. The calibration-conditioned sharpness proxy served its purpose well as a pragmatic v1 solution, and having real implementation experience behind us will make it much easier to evaluate a principled CRPS-based design when we revisit it. I'd like to keep the same RFC-first approach for that work so we can carefully validate the scoring semantics before introducing them into the Trust Score.

Thanks again for the thoughtful discussions throughout these RFCs and for the high-quality contributions. I appreciate the care you've taken to keep the implementations aligned with the overall architecture of TrustLens.

PR #166, opened 2026-07-12
feat(regression): CRPS diagnostic from multi-level intervals (RFC #155)

Whatsonyourmind:
## Summary

Implements the CRPS diagnostic from RFC #155, scoped — as discussed there — to **the estimator and its validation harness**. Diagnostic-only: no Trust Score wiring, and `sharpness_skill` is left untouched so the proxy and a proper score can be compared on real models for at least a release before any deprecation discussion.

`crps_from_intervals(y_true, intervals)` computes CRPS from the **same multi-level central-interval representation** `multilevel_interval_coverage` already consumes (no new input surface), via the quantile-loss identity:

```
CRPS(F, y) = 2 ∫₀¹ pinball_α(F⁻¹(α), y) dα
```

Each central level `τ` supplies two quantiles — `F⁻¹((1−τ)/2) = lower`, `F⁻¹((1+τ)/2) = upper` — so a set of interval levels yields a quantile grid; the pinball loss is integrated over `α` (trapezoidal) per sample and averaged.

## Validation

Estimator vs. the closed-form Gaussian CRPS `σ[ω(2Φ(ω)−1) + 2φ(ω) − 1/√π]`, forecast `N(1.3, 0.8)`:

| interval levels | # quantile levels | mean rel. error |
|---|---|---|
| 3 | 6 | ~11% (up to ~17% at tail observations) |
| 9 | 18 | 2.3% |
| 19 | 38 | 0.6% |
| 49 | 98 | 0.1% |

Error falls ~quadratically with grid density (trapezoidal convergence on a smooth pinball integrand). The estimate is **grid-dependent** — coarse/narrow grids truncate the tails and bias CRPS upward — so `n_quantile_levels` and `quantile_level_span` are returned to keep the density visible, and the docstring recommends ≥ ~9 interval levels. Quantile crossing (non-nested intervals) is repaired by sorting each sample's quantiles ascending before integration.

## Tests (`TestCrpsFromIntervals`)

- closed-form Gaussian agreement on a dense grid (within 2%);
- denser grid reduces truncation bias vs. the closed form;
- sharpness orientation (tighter forecast → lower CRPS);
- point-mass on truth → CRPS ≈ 0;
- quantile-crossing rearrangement equals the sorted-forecast score;
- grid metadata; skip-on-missing; shape / lower>upper / invalid-level / empty raises.

Full local suite green (489 passed excluding the optional xgboost/lightgbm/catboost backends); `ruff check`, `ruff format --check`, and `mypy` clean on the changed files.

## Scope / follow-up

Kept deliberately narrow so the numerics are easy to review in isolation. The **Reliability/Resolution decomposition (Hersbach 2000)** — the "why did CRPS change" interpretable terms — follows as the next PR, mirroring the RFC-first, split-PR cadence used for #157/#161. Trust Score integration (if any) is deferred until the estimator's empirical behaviour is validated here, per the RFC discussion.

Refs #155.

---
*Signed off (DCO) as Luka <luka.stanisljevic@gmail.com>.*

## Summary by CodeRabbit

* **New Features**
  * Added a new regression metric to score forecasts using multi-level central prediction intervals (CRPS from intervals).
  * Returns mean/median CRPS plus quantile-grid and sampling metadata.
  * Skips scoring gracefully when intervals are missing or insufficient for integration.

* **Bug Fixes**
  * Validates interval bounds, shapes, and level inputs; raises clear errors for invalid inputs.
  * Improves robustness by handling quantile crossing and scoring collapsed intervals near zero when forecasts match.

* **Tests**
  * Added regression coverage for correctness, edge cases, and error handling.

### DIFF
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -11,6 +11,7 @@ from sklearn.linear_model import LinearRegression
 
 from trustlens.metrics import regression
 from trustlens.metrics.regression import (
+    crps_from_intervals,
     error_distribution,
     error_variance_correlation,
     multilevel_interval_coverage,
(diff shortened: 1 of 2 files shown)


Whatsonyourmind:
Folded in the two automated-review suggestions:

- added an explicit skip test for an **empty** `intervals` dict (previously only `None` was covered) — guards the `if not intervals` gate against a future refactor to `if intervals is None`;
- documented why distinct central levels can't collide on a rounded quantile key: `(1−τ)/2 < 0.5 < (1+τ)/2`, both strictly monotone in τ, so the `round(…, 10)` only stabilises the float key and can't silently overwrite another level's quantiles. Left the code as-is since it's provably collision-free — just made the invariant explicit.

CodeRabbit came back with no actionable comments. Ready when you are.

Khanz9664 (approved):
@Whatsonyourmind 

Excellent work on this PR! I've gone through the implementation, the validation methodology, the automated review comments, and the follow-up commit, and I'm happy with where this landed.

* Implementing CRPS through the quantile-loss identity on the existing multi-level interval representation is a natural extension of the current regression diagnostics. It introduces no new input surface while providing a proper scoring rule that complements the existing interval calibration and `sharpness_skill` diagnostics.
* The numerical validation against the closed-form Gaussian CRPS is exactly the kind of evidence I like to see for statistical algorithms. Demonstrating convergence across increasing grid densities provides strong confidence that the estimator behaves as expected, while documenting the grid-dependent approximation error makes the limitations explicit rather than hidden.
* The handling of quantile crossing is a nice defensive design choice. Sorting the quantiles before integration guarantees a valid monotone inverse CDF without complicating the estimator.
* The function has comprehensive validation for malformed inputs (shape mismatches, invalid interval levels, inverted bounds, empty inputs, and insufficient quantile levels), and the graceful `"skipped"` path remains consistent with the rest of the regression metrics.

Regarding the automated review feedback:

* The additional `test_skips_when_empty_intervals_dict` is a worthwhile regression guard and addresses the coverage gap nicely.
* I also agree with the decision **not** to adopt the suggested tolerance-based merging of rounded quantile keys. The mathematical invariant documented in the code is sufficient here: lower quantiles and upper quantiles occupy disjoint regions of `(0, 1)`, and within each branch the mapping from `τ` to `α` is strictly monotonic. The `round(..., 10)` exists purely to stabilize floating-point dictionary keys, not to deduplicate mathematically equivalent quantiles. Adding a more complicated merge routine would increase complexity without providing any practical benefit.

The test suite is comprehensive as well. Beyond the standard validation paths, I especially liked seeing:

* closed-form Gaussian agreement,
* convergence with denser grids,
* orientation (sharper forecasts → lower CRPS),
* point-mass behavior,
* quantile-crossing robustness,
* metadata validation,
* and explicit edge-case handling.

Overall, this keeps the scope exactly where we discussed during the RFC: **diagnostic-only**, leaving `sharpness_skill` untouched and deferring any Trust Score integration until we have empirical experience with the metric. That separation makes the implementation much easier to reason about and review.

I don't have any further requested changes from my side. Nice work on another mathematically rigorous addition to the regression metrics.

Whatsonyourmind:
Thanks for the thorough review — and for engaging with the specific design calls, the quantile-crossing sort and the `round(…, 10)` invariant especially. Glad the validation-first framing landed.

Once this merges I'll open the reliability/resolution decomposition as PR 2/2, built on the merged estimator and held to the same bar: closed-form-anchored validation, diagnostic-only, `sharpness_skill` untouched. One wrinkle worth flagging early — the Hersbach split is a *dataset-level* property, so it'll come with its own harness (reliability ≥ 0; the decomposition summing back to CRPS on the same grid) rather than reusing the per-sample tests here. I'll keep it equally scoped and reviewable.

Khanz9664:
@Whatsonyourmind 

Thanks again for another excellent contribution!

I pulled the changes locally and did another round of verification beyond the review itself. I reran the full `tests/test_regression.py` suite and also ran a separate maintainer-level sanity check covering the estimator against the closed-form Gaussian CRPS, quantile-crossing repair, degenerate intervals, skip paths (`None` and `{}`), and some larger-scale performance checks. Everything behaved exactly as expected.

I'm particularly happy with how this landed: the estimator is mathematically well validated, the implementation is robust to edge cases, and keeping it diagnostic-only while leaving `sharpness_skill` untouched was the right design decision. It gives us a proper scoring rule without changing any existing Trust Score semantics.

I also appreciate the way you've been splitting these larger features into focused, reviewable PRs. That approach has made the RFC discussions much easier to translate into implementation while keeping each review manageable.

The decomposition PR sounds like a sensible next step. Since it's fundamentally a dataset-level property rather than a per-sample metric, keeping it in its own PR with its own validation harness (as you described) is exactly the level of separation I was hoping for. Looking forward to reviewing it when it's ready.

Thanks again for all the thoughtful work you've put into the regression uncertainty roadmap.

PR #167, opened 2026-07-13
feat(regression): CRPS reliability/resolution decomposition (RFC #155, PR 2/2)

### LINKED ISSUE #155: RFC: Interval-calibration extensions — per-quantile coverage + CRPS (follow-up to #145 / v0.5.0)

Whatsonyourmind:
## Summary

PR 2/2 for RFC #155, following the merged estimator in #166. Adds `crps_decomposition` — the Hersbach (2000) split of the CRPS into three interpretable terms, computed on the **same** multi-level interval grid `crps_from_intervals` uses:

```
CRPS = Reliability − Resolution + Uncertainty
```

- **Reliability** (≥ 0) — the calibration miss (how far stated exceedance probabilities are from observed frequencies). 0 when perfectly calibrated on the grid.
- **Resolution** (higher = better) — skill over climatology: `Uncertainty − crps_potential`.
- **Uncertainty** — CRPS of the climatological forecast (default: the marginal empirical quantiles of `y_true`; overridable via `climatology=`). A property of the data, not the model.

Diagnostic-only; `sharpness_skill` untouched; no Trust Score wiring — same discipline as #166.

## How it's estimated

Adjacent quantiles define intervals with forecast CDF level `p = α`; averaging the below/above widths across samples gives `reliability` and `crps_potential` with the **exact identity** `crps == reliability + crps_potential` (holds to machine precision by construction). `uncertainty` is the same construction applied to the climatological forecast, and `resolution = uncertainty − crps_potential`.

The reported `crps` is the decomposition-consistent (step-CDF) reconstruction; it agrees with `crps_from_intervals`' trapezoidal `mean_crps` to within the grid-interpolation gap (~3% at 19 levels, tightening as the grid densifies). Reliability/resolution are **dataset-level** properties (they need the joint distribution over many forecasts), so this is deliberately a separate function from the per-sample estimator rather than a flag on it.

## Validation (`TestCrpsDecomposition`)

Empirically verified on a wide-signal conditional Gaussian (`obs ~ N(mu,1)`, forecast `N(mu,sd)`):

| property | calibrated (sd=1) | overconfident (sd=0.5) |
|---|---|---|
| reliability | ~0.0004 (approx 0) | ~0.034 (76x higher) |
| resolution | > 0 (resolves the wide marginal) | — |
| identity `rel + pot = crps` | exact (machine precision) | exact |
| `crps` vs estimator `mean_crps` | within ~3% | — |
| climatology forecast -> resolution | ~0 | — |

Tests cover: the `reliability + crps_potential == crps` identity, the three-term identity, cross-check against the estimator, reliability >= 0 and small when calibrated / rising when overconfident, resolution > 0 for a skillful forecast and ~0 for climatology, mismatched-climatology-levels guard, and input validation.

## Refactor

Extracts the shared interval->quantile-grid validation (`_interval_quantile_grid`) now that a second metric consumes it; `crps_from_intervals` is switched to the helper with no behaviour change (its full test suite from #166 still passes). Full local suite green (501 passed, excluding the optional xgboost/lightgbm/catboost backends); `ruff`, `ruff format --check`, and `mypy` clean.

Refs #155. Follows #166.

---
*Signed off (DCO) as Luka <luka.stanisljevic@gmail.com>.*

## Summary by CodeRabbit

- **New Features**
  - Added a new CRPS decomposition metric for interval forecasts, breaking results into reliability, resolution, and uncertainty (plus related CRPS components).
  - Supports optional climatology comparisons, requiring matching interval/quantile levels for valid comparison.
  - Includes additional input validation for interval definitions and forecast/target compatibility.

- **Tests**
  - Added regression tests covering decomposition identities, behavior under (mis)calibration, agreement on dense interval grids, handling of missing intervals, and error cases for invalid or mismatched interval inputs.

### DIFF
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -403,9 +403,14 @@ class TestCrpsDecomposition:
         assert d["resolution"] > 0.0
         assert d["uncertainty"] > d["crps_potential"]
 
-    def test_climatology_forecast_has_near_zero_resolution(self):
+    def test_climatology_forecast_has_zero_resolution(self):
         # A forecast equal to climatology (marginal quantiles, constant across
-        # samples) has no skill over climatology -> resolution ~ 0.
+        # samples) has no skill over climatology -> resolution is exactly 0:
+        # resolution = CRPS_pot_clim - CRPS_pot_forecast, and the two potential
+        # CRPS terms are computed from identical quantiles. (Uncertainty must be
+        # the climatology's *potential* CRPS, not its reconstructed CRPS, for this
+        # to hold; the reconstructed variant leaks the climatology reliability
+        # into resolution.)
         rng = np.random.default_rng(6)
         y = rng.normal(0.0, 3.0, 6000)
         levels = np.round(np.arange(0.05, 0.96, 0.05), 3)
@@ -415,7 +420,7 @@ class TestCrpsDecomposition:
             hi = float(np.quantile(y, (1 + tau) / 2))
             clim[float(tau)] = (np.full(y.size, lo), np.full(y.size, hi))
         d = crps_decomposition(y, clim)
-        assert abs(d["resolution"]) < 0.02
+        assert abs(d["resolution"]) < 1e-9
 
     def test_custom_climatology_wrong_levels_raises(self):
         rng = np.random.default_rng(7)
--- a/tests/test_regression.py
+++ b/tests/test_regression.py
@@ -11,6 +11,7 @@ from sklearn.linear_model import LinearRegression
 
 from trustlens.metrics import regression
 from trustlens.metrics.regression import (
+    crps_decomposition,
     crps_from_intervals,
     error_distribution,
     error_variance_correlation,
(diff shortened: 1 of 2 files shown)


Khanz9664 (commented):
@Whatsonyourmind 

Overall this is another very high-quality PR. The refactoring into `_interval_quantile_grid()` is the right architectural move, the documentation is excellent, and the validation suite gives a lot of confidence in the implementation. I also don't believe the `round(..., 10)` concern raised by LlamaPReview is an issue—the monotonic mapping from central interval levels to lower/upper quantiles means distinct levels cannot collide, so the added documentation is sufficient.

I do have one concern before merging.

In `crps_decomposition()`, uncertainty is currently obtained via:

```python
_, _, uncertainty = _hersbach_terms(...)
```

but `_hersbach_terms()` returns:

```python
(reliability, crps_potential, crps_recon)
```

This means `uncertainty` is currently using the reconstructed climatological CRPS rather than the climatological **potential** CRPS.

Since later we compute

```python
resolution = uncertainty - crps_potential
```

a climatological forecast no longer reduces exactly to zero resolution—it instead reduces to the climatology's reliability term (which happens to be small, but not mathematically zero). The current regression test (`abs(resolution) < 0.02`) is therefore permissive enough to mask this.

Unless I'm missing something in the Hersbach formulation, I think uncertainty should come from the second return value instead:

```python
_, uncertainty, _ = _hersbach_terms(...)
```

and the climatology regression test should be tightened accordingly.

If there's a reason the reconstructed CRPS is intentionally being used here, I'd appreciate a reference or explanation, but otherwise I think this should be corrected before merging.

Whatsonyourmind:
Good catch — you're exactly right, and it's fixed in `bb53e8a`.

`_hersbach_terms()` returns `(reliability, crps_potential, crps_recon)` with `crps_recon = reliability + crps_potential`. The climatology call was taking the 3rd term (`crps_recon`), so `uncertainty` carried the climatology's own reliability, and since `resolution = uncertainty − crps_potential`, that reliability leaked straight into resolution. For a climatological forecast that left `resolution = reliability_clim` (~0.002 on the regression grid) instead of 0 — no intentional reason, just the wrong tuple slot.

Switched to the 2nd term (potential CRPS of the climatology):

```python
_, uncertainty, _ = _hersbach_terms(y_true, alphas, clim_quantiles)
```

Now `resolution = CRPS_pot_clim − CRPS_pot_forecast`, which is the standard Hersbach resolution and reduces to **exactly** 0 for a climatological forecast (the two potential terms are computed from identical quantiles). Confirmed on the regression case: resolution goes `0.002 → 0.0`, while `crps` (1.6427) still equals `uncertainty` (1.6407) + `reliability` (0.002) — so the total and reliability are unchanged; it was purely a UNC/RESOL mis-split.

Also tightened `test_climatology_forecast_has_zero_resolution` from `< 0.02` to `< 1e-9` so the invariant is actually enforced, and updated the docstrings to say the uncertainty term is the climatology's *potential* CRPS (it equals the full CRPS only when the climatology is perfectly reliable). Full suite green.

And agreed on the `round(..., 10)` point being a non-issue — thanks for the careful read.

Khanz9664 (approved):
@Whatsonyourmind 

Thanks for the quick turnaround on this.

I reviewed the follow-up commit (`bb53e8a`) and it resolves the concern I raised.

The uncertainty term is now correctly taken from the climatology's **potential CRPS** rather than the reconstructed CRPS:

```python
_, uncertainty, _ = _hersbach_terms(...)
```

which restores the intended Hersbach decomposition semantics. As expected, a climatological forecast now yields **exactly zero resolution**, while the overall CRPS and reliability terms remain unchanged. The additional inline documentation explaining why the second return value is required also makes the implementation much easier to follow for future contributors.

I also appreciate the strengthened regression test. Tightening the assertion from `abs(resolution) < 0.02` to `abs(resolution) < 1e-9` turns this from a loose behavioral check into a true mathematical invariant, which is exactly the right direction for a metric implementation.

Regarding the remaining automated comment about `round(..., 10)`: I still don't believe that represents a real issue. As discussed previously, the mapping from central interval levels to lower/upper quantiles is strictly monotonic, so distinct interval levels cannot collide after rounding in this construction. The added documentation in `_interval_quantile_grid()` sufficiently explains that invariant.

The only remaining failing check is the Security Audit workflow, which is reporting an existing dependency vulnerability and is unrelated to the changes introduced in this PR. All functional CI, linting, typing, and test workflows have passed.

With the uncertainty term corrected and the regression strengthened, I'm happy with this implementation. **Approved.**

Khanz9664:
@Whatsonyourmind 

Thanks for the quick follow-up on this.

I pulled the merged changes into my local `main` branch and reran the regression tests along with an independent maintainer sanity check.

Everything checks out. The uncertainty term is now correctly taken from the climatology's potential CRPS, and a climatological forecast now produces a resolution of exactly `0.0` as expected. I also verified that the overall CRPS decomposition identities still hold and the full regression test suite passes without issues.

Thanks for the careful implementation and for iterating on the review feedback throughout both PRs. Splitting RFC #155 into the estimator first (#166) and the decomposition second (#167) made the numerical review much easier, and the resulting implementation is both mathematically sound and well validated.

Great work!

